PR #1, opened 2017-10-10
Refactor of Xcode to allow specification of iOS simulators 

ehanlon:
- Refactor of Xcode resource. Adds ability to specify multiple versions of iOS simulator to be installed.
- Add additional tests
- Update README
- Remove references to internal tools
- Bump version to 0.8


americanhanko:
@ehanlon Maybe we should think about squashing?

PR #3, opened 2017-10-12
remove `creates` property from Xcode install block

jazaval:
This PR addresses an issue with the helper method that determines if the requested version of Xcode is already installed. The issue stems from the following `xcode-install` outputs:

`xcversion list` (listing Xcodes available to install):
```
8
8.1
8.2
8.2.1
8.3
8.3.1
8.3.2
8.3.3
9
9.1 beta
```
`xcversion installed` (listed Xcodes already installed):
```9.0	(/Applications/Xcode.app)```

Providing `xcversion install '9.0'` fails since the string `'9'` is required, so we must make some conversions between the full semantic version our Xcode resource expects in the name property and what the `xcode-install` gem needs requested.

We originally didn't notice the failure of the helper method since checking for Xcode installation was performed by the now removed `creates xcode_path` property in the Xcode install execute block.

### CODE REVIEW

File: libraries/xcode.rb [hunk 1]
```diff
@@ -2,10 +2,24 @@ module Xcode
   module Helper
     BASE_COMMAND ||= '/usr/local/bin/xcversion'.freeze
```
americanhanko: We should extract this to a method instead:

```ruby
def xcversion_command
  '/usr/local/bin/xcversion'.freeze
end

File: libraries/xcode.rb [hunk 2]
```diff
@@ -2,10 +2,24 @@ module Xcode
...
+      installed_xcodes.include?(semantic_version)
+    end
+
+    def xcversion_version(version)
+      semantic_version = version.split('.')
+      if semantic_version.length == 2 && semantic_version.last == '0'
+        semantic_version.first
+      else
+        version
+      end
```
americanhanko: Let's split this up into two different methods to make it a bit more readable:

```ruby
def needs_shortening?(version)
  split_version = version.split('.')
  split_version.length == 2 && split_version.last == '0'
end

def xcversion_version(full_version)
  full_version.first ? needs_shortening?(full_version) : full_version
end
```

jazavah:
```
[2] pry(main)> full_version.first
NoMethodError: undefined method `first' for "9.0":String
```

I like the boolean method idea but I'd need to add another `split('.')` for `full_version` in the `xcversion_version` for correct output, which seems less elegant. I've updated the variable names for better clarity.

libraries/xcode.rb
jazavah: Done.

resources/xcode.rb
```diff
@@ -7,6 +7,14 @@
 property :path, String, default: '/Applications/Xcode.app'
 property :ios_simulators, Array
 
+semantic_version = new_resource.version.split('.')
+
+version = if semantic_version.length == 2 && semantic_version.last == '0'
```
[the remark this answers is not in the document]
jazavah: I agree, but this method's purpose already takes a bit of unfolding to understand, no need to make it more obscure.

PR #5, opened 2017-10-20
use Chef's embedded Ruby instead of system Ruby

### LINKED ISSUE #4: `xcode-install` gem cannot be installed on macOS Sierra native Ruby
Attempting to install Xcode on a Sierra system using the Xcode resource fails with the error:

```
gem_package[xcode-install] action install

================================================================================
Error executing action `install` on resource 'gem_package[xcode-install]'
================================================================================

Mixlib::ShellOut::ShellCommandFailed
------------------------------------
Expected process to exit with [0], but received '1'
---- Begin output of /usr/bin/gem install xcode-install -q --no-rdoc --no-ri -v "2.3.1" --source=https://www.rubygems.org --no-document
```

It appears that the gem `jwt` can no longer be installed on native Sierra Ruby (ruby 2.0.0p648):

```
STDERR: ERROR:  Error installing xcode-install:
       	jwt requires Ruby version >= 2.1.
```

jazaval:
The native Sierra Ruby can no longer install xcode-install due to jwt requirement of >= ruby 2.1, so we install to Chef's ruby. Fixes #4.

### DIFF
--- a/libraries/xcode.rb
+++ b/libraries/xcode.rb
@@ -1,7 +1,7 @@
 module Xcode
   module Helper
     def xcversion_command
-      '/usr/local/bin/xcversion'.freeze
+      '/opt/chef/embedded/bin/xcversion'.freeze
     end
 
     def xcode_already_installed?(semantic_version)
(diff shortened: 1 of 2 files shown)


ehanlon:
Make it so.

PR #14, opened 2017-11-28
Feature/macos user

americanhanko:
### This PR:
- **New resource**:`macos_user` represents a user on the system. Properties include:
  - `autologin` - set to `true`so that this user will log in automatically after restart
  - `admin`- set to `true` to make the user an administrator
- Provides a dedicated `macos_test` cookbook, used for testing resources or recipes.
- Refactors the `systemsetup` resource for idempotence.
- Increases idempotence for `plistbuddy` using more of the custom resource DSL
  - Changes to the `plistbuddy` resource also includes more unit tests and binary conversion
- Refactor of the popular `keep_awake.rb` recipe - now fully idempotent now using `plistbuddy` and `systemsetup` only.
  **Note**: We may want to think about deprecating `pmset` going forward as it is simply a wrapper for editing `/Library/Preferences/com.apple.PowerManagement.plist`
- Adds default attributes for timezone and NTP for use with `keep_awake.rb`
- Smoke tests in the `macos_test` cookbook pass on the following macOS versions:
  - 10.11.6
  - 10.12.6
  - 10.13


resources/systemsetup.rb
```diff
@@ -3,7 +3,7 @@
 resource_name :systemsetup
 
 property :setting, desired_state: false
-property :value, desired_state: true, coerce: proc { |m| m.to_s.capitalize }
+property :value, desired_state: true, coerce: proc { |m| m.to_s.capitalize unless m.to_s.include?('/') }
```
jazaval: Is this to fix the `set networktimeserver to Time.windows.com` behavior? I'm still seeing the capitalization in the Chef run, but the keep_awake smoke test doesn't validate it - is this because they're node attributes?

recipes/keep_awake.rb
```diff
@@ -1,41 +1,80 @@
...
+  value true
+end
+
+plistbuddy 'set system sleep timer to zero' do
+  path '/Library/Preferences/com.apple.PowerManagement.plist'
+  entry 'System Sleep Timer'
+  value 0
+end
+
+plistbuddy 'disable screensaver' do
```
jazaval: If you're not using `defaults` for this preference, it needs to be written to `"/Users/#{node['macos']['admin_user']}/Library/Preferences/com.apple.screensaver.#{Hardware_UUID}.plist"` Currently the screensaver settings are not changing in a converge on all platforms.

libraries/kcpassword.rb
```diff
@@ -0,0 +1,36 @@
+module KcpasswordHelpers
```
brownmike: What's the purpose of this top level namespace? It seems we're already mixing these in to `Chef::Recipe` and `Chef::Resource`, so the extra layer of scoping doesn't seem necessary.

resources/plistbuddy.rb
```diff
@@ -1,34 +1,38 @@
 resource_name :plistbuddy
```
brownmike: Can we call this resource `plist`? `plistbuddy` hints at implementation details that are irrelevant.

resources/plistbuddy.rb
```diff
@@ -1,34 +1,38 @@
...
+
+    converge_by "set \"#{new_resource.entry}\" to #{new_resource.value} at #{new_resource.path.split('/').last}" do
+      execute plistbuddy_command :set, new_resource.entry, new_resource.path, new_resource.value
+    end
   end
-end
 
-action :delete do
-  execute format_plistbuddy_command(:delete, new_resource.entry, new_resource.value) + ' ' + new_resource.path
+  unless new_resource.binary == false
```
brownmike: Comparing to `false` is not necessary unless `nil` is a valid value.

resources/plistbuddy.rb
brownmike: Can we call this resource `plist`? `plistbuddy` hints at implementation details that are irrelevant.

libraries/kcpassword.rb
brownmike: What's the purpose of this top level namespace? It seems we're already mixing these in to `Chef::Recipe` and `Chef::Resource`, so the extra layer of scoping doesn't seem necessary.

resources/systemsetup.rb
```diff
@@ -1,26 +1,32 @@
...
-action :run do
-  if new_resource.get
-    new_resource.get.each do |flag|
-      execute BASE_COMMAND do
-        command "#{BASE_COMMAND} -get#{flag}"
-        live_stream true
-      end
-    end
+load_current_value do |desired|
+  command_output = shell_out('/usr/sbin/systemsetup', "-get#{desired.setting}").stdout.split(': ').last.strip
```
brownmike: This is the dangerous kind of method chaining that often results in `NoMethodError`s on `nil` objects at runtime. As a general rule, we should try to adhere to the Law of Demeter by only chaining methods on objects of the same type i.e. `some_string.downcase.strip.squeeze.split('//')` is okay because every method in the chain until the last one returns another string. And strictly speaking, this is not the command output; the command output ends at `.stdout`.

resources/plistbuddy.rb
americanhanko: I thought about this, and I like the idea a lot. Thought I would want @ehanlon and @jazaval before moving forward with it. In a way, it might be nice to actually incorporate the three major tools that help users manipulate property lists programmatically (`/usr/libexec/PlistBuddy`, `/usr/bin/defaults`, and `/usr/bin/plutil`), depending on the type of Plist and/or what needs to be accomplished.

recipes/keep_awake.rb
americanhanko: If that were true, wouldn't this test fail?

resources/plistbuddy.rb
jazaval: My pipe dream is that a `preference` resource uses `/usr/libexec/PlistBuddy`, `/usr/bin/defaults`, or `/usr/bin/plutil` depending on what's required, rather than require the user to know about ByHost preferences or actual plist paths. This is very lofty, but I think changing the name to `plist` is a step in the right direction, rather than trying to equate cookbook resources with command-line utilities.

recipes/keep_awake.rb
jazaval: I don't think macOS listens to the `idleTime` entry in plist being test. 
after 20 minutes

resources/plistbuddy.rb
brownmike: @jazaval I think that's a good lofty goal. From a UX perspective, we don't care what tool is used to get the job done, we just know what behavior we want. Generically naming this resource gives us reasonable freedom to use the best tool for the job.

PR #16, opened 2017-12-05
Update Xcode helpers to match namespace standard

jazaval:
This PR updates the name of the Xcode helper module and also automatically includes it in all recipes and resources. 

I've also updated the Xcode test cookbook to run a suite specific to testing Xcode installs, and updated the default Xcode version node attribute to Xcode 9.1.

### DIFF
--- a/libraries/xcode.rb
+++ b/libraries/xcode.rb
@@ -1,5 +1,5 @@
 module Xcode
-  module Helper
+  module Helpers
     def xcversion_command
       '/opt/chef/embedded/bin/xcversion'.freeze
     end
@@ -54,3 +54,6 @@ module Xcode
     end
   end
 end
+
+Chef::Recipe.include(Xcode::Helpers)
+Chef::Resource.include(Xcode::Helpers)
--- a/libraries/xcode.rb
+++ b/libraries/xcode.rb
@@ -1,47 +1,49 @@
 include Chef::Mixin::ShellOut
 
-module Xcode
-  class << self
-    def installed?(semantic_version)
-      xcversion_output = shell_out("#{XCVersion.command} installed").stdout.split
-      installed_xcodes = xcversion_output.values_at(*xcversion_output.each_index.select(&:even?))
-      installed_xcodes.include?(semantic_version)
-    end
+module MacOS
+  module Xcode
+    class << self
+      def installed?(semantic_version)
+        xcversion_output = shell_out("#{XCVersion.command} installed").stdout.split
+        installed_xcodes = xcversion_output.values_at(*xcversion_output.each_index.select(&:even?))
+        installed_xcodes.include?(semantic_version)
+      end
 
-    def bundle_version_correct?
-      xcode_version = '/Applications/Xcode.app/Contents/version.plist CFBundleShortVersionString'
-      node['macos']['xcode']['version'] == shell_out("defaults read #{xcode_version}").stdout.strip
-    end
+      def bundle_version_correct?
+        xcode_version = '/Applications/Xcode.app/Contents/version.plist CFBundleShortVersionString'
+        node['macos']['xcode']['version'] == shell_out("defaults read #{xcode_version}").stdout.strip
+      end
 
-    module Simulator
-      class << self
-        def installed?(semantic_version)
-          available_simulator_versions.include?("#{semantic_version} Simulator (installed)")
-        end
-
-        def highest_semantic_version(major_version, simulators)
-          requirement = Gem::Dependency.new('iOS', "~> #{major_version}")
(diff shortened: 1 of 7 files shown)


libraries/xcode.rb
```diff
@@ -1,5 +1,5 @@
 module Xcode
```
americanhanko: If we're going to update name the namespace, let's match the convention of the rest of the helper modules and put `MacOS` as the top level and `XcodeHelper` as the second, where the name that comes before `Helper` is the name of the resource.

resources/xcode.rb
```diff
@@ -1,5 +1,3 @@
-include Xcode::Helper
-
 resource_name :xcode
 default_action %i(setup install_xcode install_simulators)
```
americanhanko: The latest simulators are always installed with Xcode. What's the reasoning behind making `n-1` the default behavior of the resource?

libraries/xcode.rb
```diff
@@ -54,3 +54,6 @@ def available_simulator_versions
     end
   end
 end
+
+Chef::Recipe.include(MacOS::XcodeHelpers)
```
ehanlon: Why do we need to have the Xcode helpers available in all recipes and resources?

The docs point to using an [`action_class`](https://docs.chef.io/custom_resources.html#action-class) for resource actions. Are we using these in recipes?

libraries/xcode.rb
jazaval: Done.

libraries/xcode.rb
jazaval: There's actually a refactor of this using `action_class` coming up in another branch (xcode-idempotence) - I was hoping to clear this up in the meantime. Right now some of these helper methods are used in downstream cookbooks when further tinkering with Xcode installs.

americanhanko: In order for helper methods to be used in guards, they must be made available to `Chef::Recipe`. In order for them to be used "inside" a resource (i.e. a property or part of custom resource logic), they must be included in `Chef::Resource`.

You are correct about `action_class`, but that refactor is outside the scope of this PR and upcoming in `feature/xcode-idempotence`.

resources/xcode.rb
jazaval: That's kind of outside the scope of this change - but the behavior is technically determined by the node attribute `default['macos']['xcode']['simulator']['major_version'] = %w(11 10)`. I can move that attribute to the testing cookbook at some point if you'd prefer.

libraries/xcode.rb
ehanlon: 👍 Thank you for reminding me!

libraries/xcode.rb
```diff
@@ -1,5 +1,5 @@
-module Xcode
-  module Helper
+module MacOS
+  module XcodeHelpers
```
brownmike: IMO the word `Helpers` is just needless clutter in these modules. It adds no useful information at the call site.

libraries/xcode.rb
brownmike: I personally don't like adding these methods directly to the global namespace of resources & recipes. It clutters the namespace, provides poor encapsulation, and forces the use of longer than needed method names in order to clarify intent. It's much cleaner if the call site can read something like `Xcode.installed?` rather than a global method `xcode_installed?`.

libraries/xcode.rb
jazaval: I agree - this module's scope is pretty big. Any name ideas?

americanhanko: Like @ehanlon said, the best thing we could do is to extract appropriate methods to the `Chef::Provider.action_class`. The size/scope of this module will also be reduced in the `feature/xcode-idempotence` branch.

resources/xcode.rb
americanhanko: Indeed - it's outside the scope. A separate PR should be opened for that change.

libraries/xcode.rb
jazaval: I like @brownmike's suggestion of using a module's static method - `Xcode.installed?` rather than `xcode_installed`.

americanhanko: Since this is a pretty high-level library cookbook, I'd prefer we stick with idiomatic Chef and make contributions to Chef core if we think their current design modal can be improved.

jazaval: https://docs.chef.io/libraries.html uses the following custom namespace example for an `ISP` class:

```
ISP.vhosts.each do |vhost|
  directory vhost[:documentroot] do
    owner 'vhost[:uid]'
    group 'vhost[:gid]'
    mode '0755'
    action :create
  end
```
This change would put us more in line with the Chef docs, in my opinion.

PR #23, opened 2017-12-14
Improve README/Documentation

americanhanko:
This PR:

- Adds build status to the README for both 10.12 and 10.13
- Adds documentation for all of the included recipes
- Adds documentation for **ard**, **xcode**, and **plist** resources

Why:
- "RTFM"


documentation/resource_ard.md
```diff
@@ -0,0 +1,133 @@
...
+**Note:** Not all properties are compatible with each action.
+
+Actions
+-------
+
+This resource has the following actions:
+
+`:activate`
+
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Activate the remote desktop agent.
```
brownmike: Let's avoid HTML codes. They reduce the plaintext readability of markdown.

americanhanko: It was either this or write the README in reStructuredText. Markdown doesn't support indentation.

brownmike: That's a false dichotomy. We can also format the data differently so it reads well in plaintext and in a markdown renderer. I'd argue that we are displaying tabular data here, so a table can be appropriate as well.

PR #26, opened 2017-12-15
Add new property fullname to macos_user resource

mjmerin:
1. Added fullname property to macos_user resource
2. Modified test recipe for user johnny to have full name
3. Added smoke test to test for RealName using dscl command

### DIFF
--- a/resources/macos_user.rb
+++ b/resources/macos_user.rb
@@ -28,7 +28,15 @@ action_class do
 
   def full_name?
     if property_is_set?(:fullname)
-      "-fullName #{new_resource.fullname}"
+      '-fullName'
+    else
+      ''
+    end
+  end
+
+  def full_name
+    if property_is_set?(:fullname)
+      new_resource.fullname
     else
       ''
     end
@@ -45,7 +53,7 @@ end
 
 action :create do
   execute "add user #{new_resource.username}" do
-    command "#{sysadminctl} -addUser #{new_resource.username} #{full_name?} -password #{new_resource.password} #{admin_user?}"
+    command [sysadminctl, '-addUser', new_resource.username, full_name?, full_name, '-password', new_resource.password, admin_user?]
     not_if { ::File.exist? user_home }
   end
 
--- a/resources/macos_user.rb
+++ b/resources/macos_user.rb
@@ -5,6 +5,7 @@ property :username, String, name_property: true
 property :password, String, default: 'password'
 property :autologin, [TrueClass]
 property :admin, [TrueClass]
+property :fullname, String
 
 action_class do
   def user_home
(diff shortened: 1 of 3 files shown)


test/cookbooks/macos_test/recipes/new_users.rb
```diff
@@ -7,5 +7,6 @@
 
 macos_user 'create non-admin user johnny' do
   username 'johnny'
+  fullname 'Johnny\ Appleseed'
```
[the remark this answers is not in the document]
americanhanko: Should be fixable with [`Shellwords#shellescape`](https://ruby-doc.org/stdlib-2.3.0/libdoc/shellwords/rdoc/Shellwords.html#method-c-shellescape):

```ruby
argv = "It's better to give than to receive".shellescape
argv #=> "It\\'s\\ better\\ to\\ give\\ than\\ to\\ receive"
```

resources/macos_user.rb
```diff
@@ -25,6 +26,14 @@ def sysadminctl
     '/usr/sbin/sysadminctl'
   end
 
+  def full_name?
```
brownmike: A method ending with a `?` in ruby is a signal that the method returns a boolean value. It's returning a string here, so we can remove the question mark.

test/cookbooks/macos_test/recipes/new_users.rb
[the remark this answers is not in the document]
brownmike: If this is being passed to the `command` method in an `execute` resource, we can pass `command` an array of strings and it will automatically escape them for us. We shouldn't need to manually escape any strings ourselves; let the tools do that job for us.

resources/macos_user.rb
```diff
@@ -36,7 +45,7 @@ def admin_user?
 
 action :create do
   execute "add user #{new_resource.username}" do
-    command "#{sysadminctl} -addUser #{new_resource.username} -password #{new_resource.password} #{admin_user?}"
+    command "#{sysadminctl} -addUser #{new_resource.username} #{full_name?} -password #{new_resource.password} #{admin_user?}"
```
brownmike: Per another comment of mine, let's pass an array of arguments here so the `command` method can shell-escape all the strings for us.

```ruby
command [sysadminctl, '-addUser', new_resources.username, full_name, '-password', new_resource.password, admin_user?]
```

test/cookbooks/macos_test/test/smoke/default/new_users_test.rb
```diff
@@ -14,4 +14,10 @@
     its('gid') { should eq 20 }
     its('home') { should eq '/Users/johnny' }
   end
+
+  realname_cmd = 'dscl . read /Users/johnny RealName | grep -v RealName | cut -c 2-'
+
+  describe command(realname_cmd.to_s) do
```
brownmike: `realname_cmd` is already a string, we don't need to call `to_s` on it.

PR #27, opened 2017-12-19
Spec tests for XCVersion and Simulator

jazaval:
This PR adds some spec tests for the the XCVersion module and Simulator class.

### DIFF
--- a/spec/unit/libraries/xcode_spec.rb
+++ b/spec/unit/libraries/xcode_spec.rb
@@ -1,0 +1,47 @@
+require 'spec_helper'
+include MacOS::Xcode
+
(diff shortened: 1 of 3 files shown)


spec/unit/libraries/xcode_spec.rb
```diff
@@ -0,0 +1,55 @@
+require 'spec_helper'
+include MacOS::Xcode
+
+describe MacOS::Xcode::Simulator do
+  context 'when provided an available list of simulators' do
```
americanhanko: Do we need to have a similar test for testing if a particular simulator is installed?

PR #28, opened 2017-12-19
Release machine_name "fullName" property, XCVersion and Simulators unit tests

### LINKED ISSUE #25: macos_user needs fullName property
sysadminctl adduser has fullName as an option. This can be helpful when specifying a more verbose name other than username. 

`sysadminctl
-deleteUser [-secure || -keepHome]
-newPassword -oldPassword [-passwordHint ]
-resetPasswordFor -newPassword [-passwordHint ]
-addUser [-fullName ] [-UID ] [-password ] [-hint ] [-home ] [-admin] [-picture ]`

jazaval:
release #26 and #27, fixes #25


resources/macos_user.rb
```diff
@@ -25,6 +26,22 @@ def sysadminctl
     '/usr/sbin/sysadminctl'
   end
 
+  def full_name?
```
brownmike: This method is not a boolean, remove the `?`.

mjmerin: What would be a good function name for a function returns whether "-fullName" is used or not? Also should we make the same change for the admin? routine since that also returns a string.

americanhanko: We can put the logic into a single ternary statement:

```ruby
  def user_fullname
    property_is_set?(:fullname) ? "-fullName #{new_resource.fullname}" : ''
  end
```

And implement it in our execute block like this:

```ruby
  execute "add user #{new_resource.username}" do
    command [sysadminctl, '-addUser', new_resource.username, user_fullname, '-password', new_resource.password, admin_user?]
    not_if { ::File.exist?(user_home) && user_already_exists? }
  end
```

Where `user_already_exists?` is an abstraction of the guard that is used in the `:delete` action:

```ruby
  def user_already_exists?
    users = shell_out!('/usr/bin/dscl', '.', '-list', '/users').stdout
    return true if users.include?(new_resource.username)
  end
```

PR #34, opened 2018-01-26
Implement spotlight resource, metadata_util library, and add tests/documentation

jazaval:
This PR implements a `spotlight` resource, a `metadata_util` library, adds unit tests, a recipe to the test cookbook with smoke, and documentation. 

spotlight
===

Use the **spotlight** resource to manage the metadata indexing state for disk volumes. This
will primarily affect the ability to search volume contents with the macOS Spotlight feature.
Under the hood, a **spotlight** resource executes the `mdutil` command in the `metadata_util`
library.

[Learn more about Spotlight](https://support.apple.com/en-us/HT204014).

[Learn more about the `mdutil` command](https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man1/mdutil.1.html).

Syntax
------

The most basic usage of a **spotlight** resource block declares a disk volume as the name property
to **enable** metadata indexing:

```ruby
spotlight '/'
```

The full syntax for all of the properties available to the **spotlight** resource is:

```ruby
spotlight 'volume name' do
  volume                      String # defaults to 'volume name' if not specified
  indexed                     TrueClass, FalseClass # defaults to True if not specified
  searchable                  TrueClass, FalseClass # defaults to True if not specified
end
```

Actions
-------

This resource has the following actions:

`:set`

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set the metadata indexing state declared by the `indexed` property. This is the only, and default, action.

Properties
----------

`volume`

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;**Ruby Type:** `String`

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The name of the disk volume to manage.

`indexed`

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;**Ruby type:** `TrueClass, FalseClass`

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Whether or not the desired state of the named disk volume is to be indexed.

`searchable`

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;**Ruby type:** `TrueClass, FalseClass`

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Disables Spotlight searching if the index has already been created for the volume. Only applicable if the `indexed` property is set to `false`.

Examples
----------

```ruby
spotlight '/' # enables indexing on the boot volume

spotlight 'test_disk1' do # disables indexing on 'test_disk1'
  indexed false
end

spotlight 'enable indexing on TDD2' do
  volume 'TDD2'
  indexed true
end

spotlight 'disable indexing and prevent searching index on TDD-ROM' do
  volume 'TDD-ROM'
  indexed false
  searchable false
end
```
#


documentation/resource_spotlight.md
```diff
@@ -0,0 +1,87 @@
...
+```ruby
+spotlight '/'
+```
+
+The full syntax for all of the properties available to the **spotlight** resource is:
+
+```ruby
+spotlight 'volume name' do
+  volume                      String # defaults to 'volume name' if not specified
+  indexed                     TrueClass, FalseClass # defaults to TrueClass if not specified
```
americanhanko: `indexing` feels a little vague and doesn't immediately strike me as a boolean property. I think making it explicit might make it clearer, especially if it has a default property. How about something like `indexing_disabled` and have it default to `false`?

documentation/resource_spotlight.md
```diff
@@ -0,0 +1,87 @@
...
+  allow_search                TrueClass, FalseClass # defaults to TrueClass if not specified
```
americanhanko: Similar here: I think it might be better to have this be `searching_disabled` and match the same sort of format for the boolean property.

libraries/metadata_util.rb
```diff
@@ -0,0 +1,24 @@
...
+                                 'Indexing disabled.' => ['off', ''],
+                                 'Indexing and searching disabled.' => ['off', '-d'] }
+
+      @status_flags = mdutil_possible_states[volume_current_state(volume)]
+                      .insert(1, volume)
+    end
+
+    def volume_current_state(volume)
+      shell_out("/usr/bin/mdutil -s #{volume.shellescape}")
+        .stdout.split(':')[1].strip
```
americanhanko: This is a nitpick, but we do want to be careful of method chaining; as well as taking advantage of `shell_out`'s ability to take any number of arguments for its constructor (doing so _may_ allow use to avoid needing `shellescape`?)

```ruby
def volume_current_state(volume)
  status_output = shell_out('/usr/bin/mdutil', '-s', volume.shellescape).stdout.strip # handle the String first
  status_output.split(':')[1] # and the Array second
end
```

resources/spotlight.rb
```diff
@@ -0,0 +1,43 @@
...
+    ['/usr/bin/mdutil']
+  end
+
+  def desired_spotlight_state
+    [state, target_volume, search]
+  end
+end
+
+action :set do
+  execute "turn Spotlight indexing #{state} for #{target_volume}" do
```
americanhanko: > `"turn Spotlight indexing #{state} for #{target_volume}"`

Beautiful!

documentation/resource_spotlight.md
ehanlon: IMO this is fine. In my head, I usually read these properties as is it`indexed?`.

documentation/resource_spotlight.md
jazaval: I've updated this to `searchable` for clarity.

libraries/metadata_util.rb
jazaval: I've updated this method to let `shell_out` handle escape characters. Who knew? ;-)

documentation/resource_spotlight.md
jazaval: @americanhanko `indexing` isn't the property name - it's `indexed`.

americanhanko: Oops! Thanks for that correction.

What I'm getting at is that by naming the property `indexed`, it's as if there is a question mark there: `indexed?` -- that would make sense given the ruby convention for methods that return a boolean value. However, since we can't use a question mark in a property name, I thought it would be better to make it explicit: `indexing_enabled` with a default value of `false` (or vice versa).

Conversely, it turns out boolean properties across Chef (core, community or otherwise) resources do not share a consistent naming strategy, and in fact follow a similar pattern of "leaving out the question mark". Either way works for me 👍

resources/spotlight.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  def search
+    new_resource.searchable ? '' : '-d'
+  end
+
+  def target_volume
+    volume_path(new_resource.volume)
+  end
+
+  def volume_path(volume)
```
americanhanko: We can use a ternary operator here, as well as using `::File.join` to avoid the string interpolation.

```ruby
def volume_path(volume)
  volume == '/' ? volume : ::File.join '/Volumes', volume
end
```

documentation/resource_spotlight.md
```diff
@@ -0,0 +1,87 @@
...
+`:set`
+
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set the metadata indexing state declared by the `indexed`
+property. This is the only, and default, action.
+
+
+Properties
+----------
+
+`volume`
```
brownmike: IMO I think this data would look more readable in plaintext with the use of a markdown table while still being quite digestible when rendered.

PR #37, opened 2018-02-02
keep_awake.rb idempotence, improve power management resources

americanhanko:
## This PR

- Renames the `systemsetup` resource to `system_preferences`
  - This is mostly due to the description in the [`systemsetup` manual](https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man8/systemsetup.8.html)
- Adds attributes for controlling remote login and disk sleep
- Adds a library file with methods for the `keep_awake.rb` guards
- Update unit tests for High Sierra (opportunistic)
- Makes the `machine_name.rb` test recipe idempotent (a future PR will make the resource idempotent)
- Adds multi-converge and idempotence requirements to `.kitchen.yml`

## Why

- Idempotent resources will allow us to implement `why_run?` mode in the future
- Strive toward knowing that updated resources _always_ indicates a state change
- The `plist` resource was giving us false positives for power settings since any key/value pair can be written using it

## Notes
- We need a lot more unit test coverage
- More PRs are coming down the pipe in the march toward 100% idempotence
- Implementing a `launchd`/`caffeinate` recipe seems like a good alternative to offer for `keep_awake`
- Documentation is needed for `system_preferences`


recipes/keep_awake.rb
```diff
@@ -1,80 +1,72 @@
...
+  not_if { running_in_a_vm? }
 end
 
-systemsetup 'Set amount of idle time until computer sleeps to never' do
-  setting 'computersleep'
-  value 'Never'
+system_preference 'pressing power button does not sleep computer' do
+  preference :allowpowerbuttontosleepcomputer
+  setting 'Never'
+  only_if { power_button_model? }
```
ehanlon: What is the reason for restricting it to certain models? This seems like something that would be included in a cookbook that wraps around this.

test/cookbooks/macos_test/test/smoke/default/keep_awake_test.rb
```diff
@@ -28,7 +28,7 @@
     its('stdout') { should match(/ttyskeepawake\s*1/) }
   end
 
-  describe command("/usr/libexec/PlistBuddy -c 'Print :DarkWakeBackgroundTasks' '/Library/Preferences/com.apple.PowerManagement.plist'") do
+  describe command("/usr/libexec/PlistBuddy -c 'Print :DarkWakeBackgroundTasks' '/Library/Preferences/com.apple.PowerManagement.plist'"), :skip do
```
ehanlon: Could we just remove the test since this is no longer used in `keep_awake.rb`?

libraries/systemsetup.rb
```diff
@@ -0,0 +1,29 @@
...
+    def screensaver_disabled?
+      screensaver_settings('read', '0') &&
+        screensaver_settings('read-type', 'integer')
+    end
+
+    private
+
+    def screensaver_settings(query_type, expected_value)
+      expression_pattern = Regexp.new(expected_value)
+      shell_out('defaults', '-currentHost', query_type, 'com.apple.screensaver', 'idleTime',
```
jazaval: This should be it's own method so we can write unit tests for `screensaver_settings` more easily.

recipes/keep_awake.rb
```diff
@@ -1,80 +1,72 @@
-plist 'disable Power Nap' do
-  path '/Library/Preferences/com.apple.PowerManagement.plist'
-  entry 'DarkWakeBackgroundTasks'
-  value false
+system_preference 'disable computer sleep' do
+  preference :computersleep
+  setting 'Never'
```
jazaval: We should compile a large table / mapping inside `system_preference` that assigns boolean values to preferences instead of requiring users to know usage details such as "On"/"Off" vs "Never", etc. However, this is why things like `keep_awake` exist as a recipe 👍

PR #41, opened 2018-02-09
Feature/useraddtogroup: macos_user new property "groups"

mjmerin:
Add new property called "groups" to macos_user resource. This property will take either a single group in the form of a String or a list of groups in the form of an Array. 

#### Example
```
macos_user 'create new user Sam' do
    username 'sam'
    groups 'samsgroup'
end
```
#### Testing
- Add new macos_user "paul" not part of any group
- Add macos_user "randall" to the "beta" group
- Add macos_user "johnny" to the "alpha" and "beta" groups

### DIFF
--- a/resources/macos_user.rb
+++ b/resources/macos_user.rb
@@ -6,6 +6,7 @@ property :password, String, default: 'password'
 property :autologin, [TrueClass]
 property :admin, [TrueClass]
 property :fullname, String
+property :groups, [Array, String]
 
 action_class do
   def user_home
@@ -74,6 +75,24 @@ action :create do
       owner 'root'
       group 'wheel'
       mode '0600'
+    end
+  end
+
+  if property_is_set?(:groups)
+    if groups.is_a? String
+      group groups do
+        action :create
+        members username
+        append true
+      end
+    else
+      groups.each do |g|
+        group g do
+          action :create
+          members username
+          append true
+        end
+      end
     end
   end
 end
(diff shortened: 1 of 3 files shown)


resources/macos_user.rb
```diff
@@ -76,6 +77,24 @@ def user_already_exists?
...
+  if property_is_set?(:groups)
+    if groups.is_a? String
+      group groups do
+        action :create
+        members username
+        append true
+      end
+    else
+      groups.each do |g|
+        group g do
```
brownmike: What happens if the array contains a non-string? Maybe a number or a ruby object?

resources/macos_user.rb
```diff
@@ -6,6 +6,7 @@
 property :autologin, [TrueClass]
 property :admin, [TrueClass]
 property :fullname, String
+property :groups, [Array, String]
```
brownmike: I have a small nit-pick with the property name here vs. what it does. If the property name is plural, that implies that it takes more than one thing. `groups admin` just doesn't seem correct from a semantic standpoint. Maybe having two distinct properties `group` and `groups` would be more clear? Thoughts?

mjmerin: hmm...how about groups but always an Array whether it has 1 item or not? 

If someone were to ask "what groups are you in?", "I'm just in 1" is actually a fine answer too!

resources/macos_user.rb
mjmerin: We can throw some type-checking in there and enforce the usage of an array of strings. Though it might make sense to to have a group called 1, 2, 3, 4..

PR #42, opened 2018-02-12
machine_name resource idempotence + documentation updates

### LINKED ISSUE #20: `machine_name` resource does not set `LocalHostName`
Re: `machine_name` resource: one thing to keep in mind is that we don't have the same sort of logic that exists when you set a machine name using System Preferences > Sharing. In other words, `LocalHostName` cannot contain particular characters (whereas `ComputerName` and `HostName` can contain any characters).

Per the comment in the `machine_name` resource:

 ```ruby
# We cannot set the LocalHostName here because it does not conform to
# the DNS standards outlined in RFC 1034 (section 3.5)
```

We need to implement unit tests and corresponding logic for the translation of something like `New_10.13.2_WashingMachine` to `New-10132-WashingMachine`. This is what System Preferences does to make sure the `LocalHostName` meets RFC 1034

### PR DESCRIPTION
**Pull request by americanhanko**
- Updates the `machine_name` resource to be idempotent
- Adds logic to the `machine_name` resource in order to resolve #20 
- Updates the main README and adds documentation for the `system_preference` and `machine_name` resources
- Adds a `.mailmap` git file to better report the repository contributors

### CODE REVIEW

File: documentation/resource_machine_name.md [hunk 1]
```diff
@@ -0,0 +1,88 @@
+machine_name
+============
+
+Use the **machine_name** resource to manage a machine's name.
+
+As defined by the `scutil` manual, an individual macOS system has three different
+types of names managed by `scutil`: `ComputerName`, LocalHostName, and HostName.
```
ehanlon: Add back ticks before `LocalHostName` and `HostName` to align with `ComputerName`.

README.md
```diff
@@ -119,6 +126,7 @@ Resources
 ---------
 
 - [ARD (Apple Remote Desktop)](./documentation/resource_ard.md)
```
jazaval: These links should be updated with the full address since they are posted on the Chef Supermarket as well.

documentation/resource_machine_name.md
jazaval: I'm actually not sure ComputerName should have them.

ehanlon: If that's the case, we should remove them from line 14 and 15 for consistency.

PR #47, opened 2018-02-16
Whitespace support for defaults and plist resources

americanhanko:
## This PR
- Adds better handling of whitespace when maniulating plists by making sure the plist file and the plist keys/entries are properly quoted and/or escaped depending on the calling resource
- `defaults` utilizes the `coerce` property method and `Shellwords.shellescape`
- `plist` simply checks for whitespace and quotes the plist accordingly to match the other logic in the resource.


resources/defaults.rb
```diff
@@ -2,7 +2,7 @@
 
 BASE_COMMAND = '/usr/bin/defaults'.freeze
 
-property :domain, String, name_property: true
+property :domain, String, name_property: true, coerce: proc { |m| m.include?(' ') ? Shellwords.shellescape(m) : m }
```
brownmike: Any reason this needs to be here rather than where we use this property? Also it seems if it's safe to `shellescape` this property when it has spaces, it's safe to `shellescape` it all the time, no?

resources/defaults.rb
```diff
@@ -21,6 +21,7 @@
               Float.to_s => "-float #{value}" }
     value = cases[value.class.to_s]
     execute BASE_COMMAND do
+      setting = "\"#{setting}\"" if setting.include?(' ')
```
brownmike: I think the condition is not needed. It's safe to always quote the setting.

resources/defaults.rb
jazaval: This block seems unnecessary; `new_resource.domain.shellescape` on line 25 would do the same thing, right?

americanhanko:
All default suites passing locally.

PR #48, opened 2018-02-16
Improve testing for macos_user

americanhanko:
## This PR

- Adds a cookbook group to the Berksfile so that smoke tests can be run in the root cookbook directoy
- Adds better test coverage for the admin property in the macos_user resource 
- Adds `new_resource.` to properties that are missing it, which eliminates deprecation warnings


americanhanko:
This testing is going to fail until we change the build definition to remove `cd`'ing into the test cookbook directory.

Issue #50, opened 2018-02-21
Create 'security' resource


mjmerin:
The security command is the command line interface to keychains and Security framework.

Installing certificates via a new resource is helpful. Currently this can only be done via an execute block with the following:

```security import ./somecert.p12 -P secretPassword```

There are several other features that the security command is capable of and can be seen in the following [documentation](https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man1/security.1.html)

mjmerin:
Fixed in release/1.7

PR #53, opened 2018-03-01
New Custom Resource: Certificate and Library: SecurityCommand

mjmerin:
This pull request fixes Issue #50. Rather than create a security resource, the task was split up to create a security back end to drive other resources such as a certificate and even a keychain resource. 

**SecurityCommand Library**
Basic library intended to encapsulate some features of the [security command](https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man1/security.1.html). As of now this will allow you to unlock a keychain, and add a certificate of the .p12 or .cer format. 

**Certificate Custom Resource**
This is a new custom resource that utilizes the SecurityCommand library mentioned above as its backend. It currently only has 3 properties for basic certificate install. 

*Notes*
- For testing, a PFX format Test.p12 has been added. This is a dummy certificate file


libraries/security_cmd.rb
```diff
@@ -0,0 +1,44 @@
...
+      @cert = cert
+      @keychain = keychain
+      @security = '/usr/bin/security'
+    end
+
+    def unlock_keychain(kc_passwd)
+      [@security, 'unlock-keychain', '-p', kc_passwd]
+    end
+
+    def add_certificates
```
americanhanko: Use a ternary statement here:
```
@keychain == '' ? [@security, 'add-certificates', @cert] : [@security, 'add-certificates', @cert, '-k', @keychain]
```

resources/certificate.rb
```diff
@@ -0,0 +1,28 @@
+resource_name :certificate
+default_action :install
+
+property :certfile, String
+property :cert_password, String
+property :keychain, String
+
+action_class do
+  def keychain
```
americanhanko: Ternary statement here as well:

```ruby
property_is_set?(:keychain) ? new_resource.keychain : ''
```

resources/certificate.rb
```diff
@@ -0,0 +1,28 @@
...
+      ''
+    end
+  end
+end
+
+action :install do
+  test = SecurityCommand.new(new_resource.certfile, keychain)
+
+  execute 'unlock keychain' do
+    command [*test.unlock_keychain('vagrant')]
```
americanhanko: Let's change `'vagrant'` to `node['macos']['admin_user']` attribute; just in case someones uses a different username in their environment.

resources/certificate.rb
```diff
@@ -0,0 +1,28 @@
...
+    if property_is_set?(:keychain)
+      new_resource.keychain
+    else
+      ''
+    end
+  end
+end
+
+action :install do
+  test = SecurityCommand.new(new_resource.certfile, keychain)
```
americanhanko: I think we can use a better name than `test` here: perhaps `certificate` or `cert`

libraries/security_cmd.rb
```diff
@@ -0,0 +1,44 @@
...
+        [@security, 'add-certificates', @cert, '-k', @keychain]
+      end
+    end
+
+    def import(cert_passwd)
+      [@security, 'import', @cert, '-P', cert_passwd, @keychain]
+    end
+
+    def install_certificate(cert_passwd)
+      if ::File.extname(@cert) == '.p12'
```
americanhanko: In the [X.509 standard](https://en.wikipedia.org/wiki/X.509#Certificate_filename_extensions), there are a few different file extensions used for the various types of files. For certificates, we might find them with `.der`, `.crt`, or `.cer` and for [ PKCS #12](https://en.wikipedia.org/wiki/PKCS_12) files, we might find them with `.p12` of `pfx`. 

Might it make sense to check for these possible variations? Could we test for that?

resources/certificate.rb
```diff
@@ -0,0 +1,28 @@
+resource_name :certificate
+default_action :install
```
americanhanko: If there is only one action, we don't need to indicate a default. If we decide to add more actions later, then we'll need to add it. I would probably remove it until then, but it's up to you.

spec/unit/libraries/security_cmd_spec.rb
```diff
@@ -0,0 +1,67 @@
+require 'spec_helper'
+include MacOS
+
+describe MacOS::SecurityCommand do
+  context 'when SecurityCommand object is instantiated' do
+    it 'security should use /usr/bin/security' do
+      test = MacOS::SecurityCommand.new('/Users/vagrant/Test.p12', '')
```
americanhanko: We don't need to instantiate this object for every test. You should be able to put it right inside the `describe` block and then it should be available to all the tests.

libraries/security_cmd.rb
mjmerin: The 'import' command can actually support the formats: openssl, bsafe, raw, pkcs7, pkcs8, pkcs12, x509, openssh1, openssh2, and pemseq. We can probably make an array of acceptable file formats and check against those.

libraries/security_cmd.rb
```diff
@@ -0,0 +1,44 @@
+include Chef::Mixin::ShellOut
+
+module MacOS
+  class SecurityCommand
+    attr_reader :cert
+    attr_reader :keychain
+    attr_reader :security
```
americanhanko: If this attribute is the path to the command, let's indicate that with the name. Something like `:security_executable`, `:security_command` or `:security_cmd`

libraries/security_cmd.rb
mjmerin: done

resources/certificate.rb
mjmerin: done

resources/certificate.rb
mjmerin: done

resources/certificate.rb
mjmerin: done

libraries/security_cmd.rb
```diff
@@ -0,0 +1,44 @@
...
+      [@security, 'import', @cert, '-P', cert_passwd, @keychain]
+    end
+
+    def install_certificate(cert_passwd)
+      if ::File.extname(@cert) == '.p12'
+        import(cert_passwd)
+      elsif ::File.extname(@cert) == '.cer'
+        add_certificates
+      else
+        Chef::Exception.fatal('invalid cert file')
```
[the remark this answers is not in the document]
mjmerin: done

resources/certificate.rb
mjmerin: done

spec/unit/libraries/security_cmd_spec.rb
mjmerin: done

libraries/security_cmd.rb
mjmerin: done

PR #59, opened 2018-03-12
Release/1.8

americanhanko:
## Release 1.8

- Add `dns_domain` property to `machine_name` resource to support FQDNs
- Remove support for `NetBIOSName` due to macOS bugs
- Add TESTING.md documentation
- Fix several bugs in `plist` resource
- Add new suite for `machine_name` test recipe
- Fix several typos in `machine_name` resource documentation
- Change `binary` property to `encoding` to support xml and binary plist formats


TESTING.md
```diff
@@ -0,0 +1,159 @@
...
+For the specific suite, there are corresponding integration tests as specified
+in the suite. Every suite is tested against all three platform version.
+
+### Building a macOS Vagrant Box
+
+Due to Apple's Software License Agreement, you'll need to build your own boxes.
+There's a number of different resources on GitHub that provide some really great
+guys, but we're partial to [osx-vm-templates](https://github.com/timsutton/osx-vm-templates).
+
+This procedure is a bit of a pain to really nail down. We've been working on
```
ehanlon: Maybe mention the @americanhanko fork of osx-vm-templates here?

TESTING.md
```diff
@@ -0,0 +1,159 @@
...
+$ kitchen test default.*101[23] # only test default suites on 10.12 and 10.13
+```
+
+macOS takes a little while to boot and the suites themselves (especially Xcode)
+can take a while to run - some of our builds end up being 30-40 minutes per operating
+system. If you've got the hardware, don't be afraid to run
+`kitchen test --concurrency n` to save a little time (where `n` is the number of concurrent
+instances you want to boot up).
+
+## Testing via a Pull Request
```
ehanlon: I don't feel that it's a good idea to offer up our infrastructure as a path to testing their changes. My concern is that we have other jobs that run on our Mac build machines, and if they're tied up running tests for someone else's change it could have an impact.

libraries/machine_name.rb
```diff
@@ -1,28 +1,53 @@
 module MacOS
   module MachineName
-    def scutil
-      '/usr/sbin/scutil'
+    def conform_to_rfc1034(hostname)
```
ehanlon: A more descriptive name? Perhaps `validate_domain_name` or `validate_hostname`?

resources/machine_name.rb
```diff
@@ -4,54 +4,73 @@
...
 action :set do
+  property_is_set?(:netbios_name) ? new_resource.netbios_name : new_resource.netbios_name = new_resource.hostname
+  property_is_set?(:computer_name) ? new_resource.computer_name : new_resource.computer_name = new_resource.hostname
+  property_is_set?(:local_hostname) ? new_resource.local_hostname : new_resource.local_hostname = new_resource.hostname
+
   converge_if_changed :hostname do
-    converge_by "set Hostname to #{new_resource.hostname}" do
-      execute [scutil, '--set', 'HostName', new_resource.hostname] do
+    converge_by 'set Hostname' do
+      fqdn = [new_resource.hostname, new_resource.dns_domain].join '.'
```
ehanlon: This appends the `HostName` with a `.` at the end if the `dns_domain` property is not set.

TESTING.md
jazaval: Let’s see what the actual utilization looks like before we decide this.

resources/machine_name.rb
americanhanko: Good catch, thanks.

TESTING.md
americanhanko: I can see both points. Right now, we don't have a lot of PRs being opened up - but that is presumably only going to increase. Ultimately, I think this section is better suited for a `CONTRIBUTING.md` though, so I'm going to remove from `TESTING.md` until that is implemented.

libraries/machine_name.rb
americanhanko: I wanted to make it specific to standardization. It's also performing an action on the entered hostname, so I chose a verb rather than a predicate method name.

How about `conform_to_dns_standards`?

ehanlon: Works for me 👍

PR #62, opened 2018-03-15
Add support for other hypervisors in keep_awake

americanhanko:
- Leverage the `node['virtualization']['systems']` ohai attribute to
determine whether we are a guest or a host
- Add appropriate ChefSpec tests for parallels guest, host, or empty hash


americanhanko:
Changing the base branch to `release/1.9` so that we have a little breathing room. It's possible we'll see changes to ohai, which may change our implementation for this feature.

tas50:
This will ship with Chef/Ohai 14 and the next release we do of Ohai/Chef 13
https://github.com/chef/ohai/pull/1164

americanhanko:
Merging and then confirming the release/1.9 builds before merging into master

PR #66, opened 2018-03-15
Resolve build failure using correct InSpec syntax

americanhanko:
There's a whitespace issue, but I'll take care of that in the next hotfix PR for `keep_awake` resources updated.


jazaval:
Resolves #66

PR #72, opened 2018-03-21
Hotfix/plist encoding

americanhanko:
Make the plist resource able to handle files that are encoded as utf-8


americanhanko:
Build failure on 10.12.6 due to keep_awake idempotence - going to complete the merge.

PR #74, opened 2018-03-21
Feature/group smoke test 67

mjmerin:
**Separated new_users recipe into its own test suite called users**
As this was a fairly hard to reproduce bug, simplifying the test allowed easier reproducability

**Add half a second sleep**
The issue would only manifest once in roughly every 10-15 kitchen test runs. It was discovered that when throwing in debug printouts, the issue would never manifest. Adding 0.5 second sleep seems to produce the same results. The test has been run over 70 times without the issue reproducing.

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -73,3 +73,10 @@ suites:
     verifier:
       inspec_tests:
         - test/smoke/certificate
+
+  - name: users
+    run_list:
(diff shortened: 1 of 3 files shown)


jazaval:
Looks good to me - thanks!

mjmerin:
Build agents changed resulting in build issues. Will close and reopen.

PR #78, opened 2018-03-24
Restructured Integration Testing Suite using InSpec Profile

americanhanko:
## This PR
- Converts the `chef generate cookbook` smoke test layout into an [InSpec profile](https://www.inspec.io/docs/reference/profiles/) layout, created by `inspec init profile`.
- Breaks up each of the InSpec tests into control groups that make a little more sense (e.g. `admin-user` vs. `standard-user` controls)

## Why
 - I feel this is a better design for a testing suite that allows the easy re-use of InSpec controls.
- With an InSpec profile, there is potential for reusability for other macOS admins/ops personnel.

## Takeaways
- The controls are named and organized according to **System Preferences.app**. Although this structure feels purposeful and familiar, it may need to be revisited if it is too consumer forward rather than administrator forward.
- Significantly more test coverage is required for things like preferences and each of their domain scopes. To combat this, I hope to add some high-level functionality via the use of attributes or recipes that will make the resources of the cookbook more useful for less technical users


.kitchen.yml
```diff
@@ -4,79 +4,72 @@ driver:
   provider: parallels
 
 provisioner:
-  product_name: chef
-  product_version: 13.8
+  name: chef_zero
```
jazaval: I think we should still be forcing the latest chef-client here, instead of what might be in a testing box.

.kitchen.yml
```diff
@@ -4,79 +4,72 @@ driver:
...
-      - recipe[macos_test::xcode]
-    verifier:
-      inspec_tests:
-        - test/smoke/xcode
+- name: spotlight
+  run_list:
+  - recipe[macos_test::spotlight]
+  verifier:
+    controls:
+    - indexing/searching
```
jazaval: These slashes make me thing there's some sort of namespacing happening at first glance - may just `indexing_and_seaching`?

.kitchen.yml
```diff
@@ -4,79 +4,72 @@ driver:
...
-      - recipe[macos_test::spotlight]
-    verifier:
-      inspec_tests:
-        - test/smoke/spotlight
+- name: xcode
+  run_list:
+  - recipe[macos_test::xcode]
+  verifier:
+    controls:
+    - xcode/simulators
```
jazaval: These slashes make me thing there's some sort of namespacing happening at first glance - may just `xcode_and_simulators`?

test/integration/default/README.md
```diff
@@ -0,0 +1,6 @@
+# macOS Cookbook InSpec Profile
+
+The macOS Cookbook InSpec profile tests the custom resources and the recipes
+that utilize them to assertain their full functionality and dependability. We
```
jazaval: *ascertain

test/integration/default/controls/desktop_and_screensaver.rb
```diff
@@ -0,0 +1,35 @@
+title 'desktop & screen saver'
```
jazaval: I would prefer `and` over `&`.

test/integration/default/controls/energy_saver.rb
```diff
@@ -0,0 +1,68 @@
...
+  title 'power management settings for managed macs'
+  desc '
+    Verify all sleep is disabled the machine automatically
+    restarts after a freeze or power outage using the systemsetup and
+    pmset commands
+  '
+
+  describe command('systemsetup -getremoteappleevents') do
+    apple_events_pattern = Regexp.union('Remote Apple Events: On')
+    its('stdout') { should match apple_events_pattern }
```
jazaval: You can use the `include` matcher to write this more simply. 
```
its('stdout') { should include 'Remote Apple Events: On' }

File: test/integration/default/controls/xcode.rb [hunk 7]
```diff
@@ -0,0 +1,39 @@
...
+    it { should be_symlink }
+  end
+
+  if macos_version.match? Regexp.union ['10.12', '10.13']
+    describe directory('/Applications/Xcode-9.2.app') do
+      it { should exist }
+    end
+
+    describe command('/opt/chef/embedded/bin/xcversion simulators') do
+      sim_pattern = Regexp.union 'iOS 10.3.1 Simulator (installed)'
```
jazaval: Consider using the `include` matcher here as well.

test/integration/default/controls/xcode.rb
```diff
@@ -0,0 +1,39 @@
...
+      its('exit_status') { should eq 0 }
+    end
+
+  elsif macos_version.match? Regexp.union '10.11'
+    describe directory('/Applications/Xcode-8.2.1.app') do
+      it { should exist }
+    end
+
+    describe command('/opt/chef/embedded/bin/xcversion simulators') do
+      sim_pattern = Regexp.union 'iOS 9.3 Simulator (installed)'
```
jazaval: And here.

test/integration/default/controls/desktop_and_screensaver.rb
americanhanko: Why I wrote it with the ampersand:

screenshot 2018-03-24 15 43 19

.kitchen.yml
jazaval:
```
provisioner:
  product_name: chef
  product_version: latest
  install_strategy: always
```

PR #79, opened 2018-03-25
Install Command Line Tools as part of Xcode setup

jazaval:
This PR ensures the `xcode-install` gem can be installed into Chef gems by verifying that Command Line Tools are installed. If not installed, a temporary file is created to trigger the `softwareupdate -l` command into providing the product name for the latest CLT, which is then downloaded and installed.

Fixes #79.

### DIFF
--- a/libraries/command_line_tools.rb
+++ b/libraries/command_line_tools.rb
@@ -1,0 +1,20 @@
+include Chef::Mixin::ShellOut
+
+module MacOS
+  class CommandLineTools
+    attr_reader :product
+
+    def initialize
+      @product = available_products.lines
+                                   .select { |s| s.include?('* Command') }
+                                   .last.tr('*', '').strip
+    end
+
+    def available_products
+      shell_out('softwareupdate -l').stdout
+    end
+  end
+end
+
+Chef::Recipe.include(MacOS)
+Chef::Resource.include(MacOS)
--- a/libraries/command_line_tools.rb
+++ b/libraries/command_line_tools.rb
@@ -17,7 +17,7 @@ module MacOS
     end
 
     def available_products
-      shell_out('softwareupdate -l').stdout
+      shell_out(['softwareupdate', '--list']).stdout
     end
   end
 end
--- a/libraries/command_line_tools.rb
+++ b/libraries/command_line_tools.rb
@@ -5,9 +5,15 @@ module MacOS
     attr_reader :product
 
     def initialize
-      @product = available_products.lines
-                                   .select { |s| s.include?('* Command') }
-                                   .last.tr('*', '').strip
+      @product = if available_command_line_tools.empty?
+                   'none'
+                 else
+                   available_command_line_tools.last.tr('*', '').strip
+                 end
+    end
+
+    def available_command_line_tools
+      available_products.lines.select { |s| s.include?('* Command') }
     end
 
     def available_products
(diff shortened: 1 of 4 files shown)


libraries/command_line_tools.rb
```diff
@@ -0,0 +1,20 @@
...
+    attr_reader :product
+
+    def initialize
+      @product = available_products.lines
+                                   .select { |s| s.include?('* Command') }
+                                   .last.tr('*', '').strip
+    end
+
+    def available_products
+      shell_out('softwareupdate -l').stdout
```
americanhanko: Nit: Break the command into separate arguments and use the long version of the `list` option?

```ruby
shell_out('softwareupdate', '--list').stdout
```

resources/xcode.rb
```diff
@@ -6,6 +6,18 @@
...
 
 action :setup do
+  file 'sentinel to make CLT available' do
+    path '/tmp/.com.apple.dt.CommandLineTools.installondemand.in-progress'
+    not_if { ::File.exist?('/Library/Developer/CommandLineTools/usr/lib/libxcrun.dylib') }
+    notifies :run, 'execute[install Xcode CLT]', :immediately
+  end
+
+  execute 'install Xcode CLT' do
+    command lazy { "softwareupdate --verbose -i '#{CommandLineTools.new}'" }
```
americanhanko: A couple legitimate questions:
- Based on the unit test, do we need to use `CommandLineTools.new.product` in the command?
- Do we want to be live-streaming here? If not, is the `verbose` option needed?

Nitpicks. Could we:
- extract the command to a list variable for readability (clearer that we need `CommandLineTools` in quotes)
- use the long version of the `install` option

```ruby
execute 'install Xcode CLT' do
  install_clt_command = ['softwareupdate', '--install', "'#{CommandLineTools.new}'"]
  command lazy { install_clt_command }
  action :nothing
  notifies :delete, 'file[sentinel to make CLT available]', :immediately
end
```

spec/unit/libraries/command_line_tools_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+describe MacOS::CommandLineTools do
+  context 'when provided an available list of software update products' do
+    before do
+      allow_any_instance_of(MacOS::CommandLineTools).to receive(:available_products)
+        .and_return(<<-SOFTWAREUPDATE_LIST
+                    Software Update Tool
+                    Finding available software
+                    Software Update found the following new or updated software:
+                       * Command Line Tools (macOS El Capitan version 10.11) for Xcode-8.2
+                    \tCommand Line Tools (macOS El Capitan version 10.11) for Xcode (8.2), 150374K [recommended]
```
americanhanko: Oh, neat. I didn't realize you could use`\t` for a tab in a heredoc!

resources/xcode.rb
jazaval: Yep.

libraries/command_line_tools.rb
jazaval: Yep.

resources/xcode.rb
jazaval: `CommandLineTools` also is handled by the command array properly (something I hadn't tested before)

spec/unit/libraries/command_line_tools_spec.rb
jazaval: I meant to use `<<~` here actually - the tabs make more sense in that context.

americanhanko:
Looks great. Thanks for making the changes quickly.

Issue #76, opened 2018-03-22
Add support for apple_id credentials via chef-vault


StefanRehder:
It would be nice if the values of XCODE_INSTALL_USER and XCODE_INSTALL_PASSWORD could be set via attributes. For example I would like to use chef-vault instead of data bags but the xcode resource currently requires a data bag.

I would like to be able to do something like this in my recipe:

```
include_recipe ‘chef-vault’

vault = chef_vault_item(:credentials, ‘apple_id’)
node.set['macos'][‘apple_id’][‘user’] = vault[‘user’]
node.set['macos'][‘apple_id’][‘password’] = vault[‘password’]

include_recipe 'macos::xcode'
```

It is fine if you default them to the data bag values but it would be nice with an option to override the default values if you want to use chef-vault for example.

americanhanko:
Thanks for this feature request, @StefanRehder - we will certainly look into making this happen.

americanhanko:
Added in #82

PR #81, opened 2018-03-26
Allow node attributes to be set for Apple ID credentials, increase install timeout

### LINKED ISSUE #80: Execute block in xcode resource exceeds the timeout limit
This execute block in the xcode resource takes more than 1 hour to complete on my machine (with ~14 Mbit/s bandwidth):

```
action :install_xcode do
  execute "install Xcode #{new_resource.version}" do
    environment DEVELOPER_CREDENTIALS
    command XCVersion.install_xcode(new_resource.version)
    not_if { Xcode.installed?(new_resource.version) }
  end
end
```

So it fails with:
```
ERROR: xcode[9.2] (macos::xcode line 1) had an error: Mixlib::ShellOut::CommandTimeout: execute[install Xcode 9.2] (/tmp/kitchen/cache/cookbooks/macos/resources/xcode.rb line 27) had an error: Mixlib::ShellOut::CommandTimeout: Command timed out after 3600s:
       Command exceeded allowed execution time, process terminated
       ---- Begin output of /opt/chef/embedded/bin/xcversion install '9.2' ----
       STDOUT:
######################################################################    97.7%######## 100.0%
       ---- End output of /opt/chef/embedded/bin/xcversion install '9.2' ----
```

It looks like it was very close to completion within the time limit though.
Is it possible to extend the [timeout property](https://docs.chef.io/resource_execute.html) of the execute block (maybe 2 hours is a better timeout limit)?

I think it is fair to assume that a 5GB download + installation will exceed 1 hour for a lot of people.

jazaval:
This PR allows node attributes to be set for Developer Apple ID credentials while downloading Xcode from Apple: 

```
node['macos']['apple_id']['user']
node['macos']['apple_id']['password']
```
This PR also increases the timeout for the Xcode download and install process to 2 hours, and deletes a library method `bundle_version_correct?` since it doesn't work (unsuccessfully tries to access node attributes in Xcode library).

Fixes #80 and #76.

### DIFF
--- a/libraries/xcode.rb
+++ b/libraries/xcode.rb
@@ -9,12 +9,14 @@ module MacOS
         installed_xcodes.include?(semantic_version)
       end
 
-      def find_apple_id(data_bag_retrieval, node_attributes)
-        data_bag_retrieval.call
+      def find_apple_id(data_bag_retrieval, node_credential_attributes)
+        if node_credential_attributes
+          { 'apple_id' => node_credential_attributes['user'],
+            'password' => node_credential_attributes['password'] }
+        else
+          data_bag_retrieval.call
+        end
       rescue Net::HTTPServerException
-        { 'apple_id' => node_attributes['user'],
-          'password' => node_attributes['password'] }
-      rescue NoMethodError
         Chef::Application.fatal!('No developer credentials supplied!')
       end
     end
--- a/libraries/xcode.rb
+++ b/libraries/xcode.rb
@@ -13,6 +13,15 @@ module MacOS
         xcode_version = '/Applications/Xcode.app/Contents/version.plist CFBundleShortVersionString'
         node['macos']['xcode']['version'] == shell_out("defaults read #{xcode_version}").stdout.strip
       end
+
+      def find_apple_id(data_bag_retrieval, node_attributes)
+        data_bag_retrieval.call
+      rescue Net::HTTPServerException
+        { 'apple_id' => node_attributes['user'],
+          'password' => node_attributes['password'] }
+      rescue NoMethodError
+        Chef::Application.fatal!('No developer credentials supplied!')
+      end
     end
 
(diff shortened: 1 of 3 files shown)


libraries/xcode.rb
```diff
@@ -9,9 +9,15 @@ def installed?(semantic_version)
         installed_xcodes.include?(semantic_version)
       end
 
-      def bundle_version_correct?
-        xcode_version = '/Applications/Xcode.app/Contents/version.plist CFBundleShortVersionString'
-        node['macos']['xcode']['version'] == shell_out("defaults read #{xcode_version}").stdout.strip
```
americanhanko: FWIW, this could still be accessed (set in this case), with `Chef.node.default['macos']['xcode']['version']`. However, since this method wasn't even used, I don't think there's any reason to keep it. Down the road, we'll probably use something like this when we implement `load_current_value` and check the current version before doing anything.

libraries/xcode.rb
```diff
@@ -9,9 +9,15 @@ def installed?(semantic_version)
...
+      def find_apple_id(data_bag_retrieval, node_credential_attributes)
+        if node_credential_attributes
+          { 'apple_id' => node_credential_attributes['user'],
+            'password' => node_credential_attributes['password'] }
+        else
+          data_bag_retrieval.call
+        end
+      rescue Net::HTTPServerException
```
americanhanko: Is there a chance that we could get this exception under any other circumstances? If so, maybe our message should say something more generic, but include a hint that it is probably missing credentials that caused the error. IIRC, we've gotten errors before due to Apple's servers not responding, but I'm not sure if that's this exact exception.

jazaval: This `rescue` will only save exceptions arising from `data_bag_retrieval.call`, which is before we ever reach out to Apple.

PR #82, opened 2018-03-26
Release/1.10

americanhanko:
### Changelog

- #66 Resolve build failure using correct InSpec syntax
- #77 Resolve issue when adding users to groups
- #78 Revamped InSpec profile
- #79 Install Command Line Tools as part of Xcode setup
- #80 Execute block in Xcode resource times out
- #81 Allow node attributes for Apple ID credentials


americanhanko:
Building failing intermittently due to issues with `keep_awake` on 10.13.3. Maybe related to the box itself or a race condition in the recipe. All other tests passing, so we're ready to merge.

PR #84, opened 2018-04-02
Support Xcode beta and GM installation

jazaval:
This PR allows the Xcode resource to support betas and GMs from Apple.

It refactors the `MacOS::Xcode` module into a Class, and in doing so, takes the handling of version semantics away from `MacOS::XCVersion` and credentials management away from the Xcode resource itself.

It adds `Xcode::Version` to handle some of the semantics Apple has with it's Xcode versioning, which removes the overloaded `apple_pseudosemantic_version` method.

It also updates the default Xcode recipe to install the latest possible Xcode for each platform, adds a `xcode-beta` test suite to Kitchen, and avoids installing Chef 14 in Kitchen.


libraries/xcode.rb
```diff
@@ -1,24 +1,54 @@
...
+      developer_id = find_apple_id(data_bag_retrieval, node_credential_attributes)
+      @credentials = {
+        XCODE_INSTALL_USER:     developer_id['apple_id'],
+        XCODE_INSTALL_PASSWORD: developer_id['password'],
+      }
+      authenticate_with_apple(@credentials)
+      @version = apple_pseudosemantic_version(semantic_version)
+    end
+
+    def apple_pseudosemantic_version(semantic_version)
```
brownmike: This method seems to be doing a lot. Furthermore, it looks like it's actually returning available Xcode versions that match the passed in version? I think? It's a bit hard to follow.

libraries/xcode.rb
```diff
@@ -1,24 +1,54 @@
...
+      split_version = semantic_version.split('.')
```
brownmike: Have we considered using `Gem::Version` objects? They're much easier to inspect and compare than strings.

libraries/xcode.rb
brownmike: Ah, it's returning the first version that matches the passed in version. Either way, this method is probably doing too much.

test/integration/default/controls/xcode.rb
```diff
@@ -35,3 +30,24 @@
...
+  '
+
+  macos_version = command('/usr/bin/sw_vers -productVersion').stdout.strip
+
+  describe file('/Applications/Xcode.app') do
+    it { should exist }
+    it { should be_symlink }
+  end
+
+  if macos_version.match? Regexp.union ['10.12', '10.13']
```
brownmike: Nit pick: the brackets are not necessary `Regexp.union '10.12', '10.13'` will work the same way.

libraries/xcode.rb
jazaval: I looked into trying this since I use `Gem` objects in the `Simulator` class. The problem I'm trying to solve is that Apple's versioning for new major releases of Xcode breaks semantic rules. The major releases have to be referred to as `Xcode 9`, not `Xcode 9.0`.  This method was built to strip the `.0` that users of the Xcode resource provide in the `version` property - I'm not sure how I can effectively utilize a `Version` object. 

Thanks for pointing out that it grew past it's original purpose though, I've split this into two methods.

libraries/xcode.rb
jazaval: Thanks, I've split this up.

libraries/xcode.rb
brownmike: Sounds like we need our own version object. Would this work?

```ruby
2.5.1 (main):0 > class XcodeVersion < Gem::Version
2.5.1 (main):0 *   def major
2.5.1 (main):0 *     _segments.first
2.5.1 (main):0 *   end
2.5.1 (main):0 *   def minor
2.5.1 (main):0 *     _segments[1] || 0
2.5.1 (main):0 *   end
2.5.1 (main):0 *   def patch
2.5.1 (main):0 *     _segments[2] || 0
2.5.1 (main):0 *   end
2.5.1 (main):0 * end
=> :patch
2.5.1 (main):0 > XcodeVersion.create '9'
=> #<XcodeVersion "9">
2.5.1 (main):0 > v = _
=> #<XcodeVersion "9">
2.5.1 (main):0 > v.major
=> 9
2.5.1 (main):0 > v.minor
=> 0
2.5.1 (main):0 > v.patch
=> 0
```

PR #88, opened 2018-04-10
Add suites for Chef 13 into Kitchen

jazaval:
This PR ensures that we don't break anything for Chef 13 users.

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -40,13 +40,13 @@ platforms:
 
 - name: high-sierra-chef13
   driver:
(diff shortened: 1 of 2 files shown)


americanhanko:
Just going off Fletcher's gist [Testing multiple versions of Chef with Test Kitchen](https://gist.github.com/fnichol/77c9ba7c64d9b6efa936) here, but could we specify the provisioner settings in the platforms and then name each platform to include the Chef version? Would help eliminate some duplication.

PR #99, opened 2018-04-21
New Keep Awake Tests

mjmerin:
- Rename systemsetup library to system to match the module name
- Add test for portable bare metal
- Add tests for using the new FormFactor class


jazaval:
The build failure isn't persisting and is part of the Xcode suite anyways - merging.

PR #106, opened 2018-04-30
Release/1.14

jazaval:
This PR updates the `keep_awake` recipe and spec tests to not require node attribute stubbing when wrapped in another cookbook. 

It also removes some extra specs tests where logic was overtested (`should_not eq`) and updates some of the spec language for clarity.

### DIFF
--- a/libraries/system.rb
+++ b/libraries/system.rb
@@ -3,18 +3,18 @@ module MacOS
     class FormFactor
       attr_reader :machine_model
 
-      def initialize(machine_model)
-        @machine_model = machine_model
+      def initialize(hardware)
+        @machine_model = hardware.nil? ? nil : hardware['machine_model']
       end
 
       def desktop?
         return false if @machine_model.nil?
-        @machine_model.match? Regexp.union %w(MacMini MacPro iMac)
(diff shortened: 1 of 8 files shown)


libraries/system.rb
```diff
@@ -3,18 +3,18 @@ module System
     class FormFactor
       attr_reader :machine_model
 
-      def initialize(machine_model)
-        @machine_model = machine_model
+      def initialize(hardware)
+        @machine_model = hardware.nil? ? nil : hardware['machine_model']
```
brownmike: Nitpick: If you use empty string instead of `nil`, you won't need the `return false if @machine_model.nil?` duplication below.

brownmike: Actually if it's possible for `node['hardware']` to be `nil` then this would fail. If that's possible, we can still remove the duplication here with something like this:

```ruby
def initialize(hardware)
  @machine_model = (hardware.nil? || !hardware['machine_model']) ? '' : hardware['machine_model']
```

jazaval: I'll wait to refactor this until we put a PR into fauxhai.

PR #110, opened 2018-05-11
Fix autologin functionality on 10.13.4

jazaval:
This PR fixes an issue where the autologin user prompted for a password the first reboot after converge. I resolved this by adding `'lastUser' => 'loggedIn'` to the loginwindow plist - the default value causing the bug was `'lastUser' => 'Restart'`.

Resolving this bug also exposed that we were not dismissing the Data & Privacy setup pane new in 10.13.4. I added `'DidSeePrivacy' => true` to the user's SetupAssistant plist to dismiss this.

### DIFF
--- a/resources/macos_user.rb
+++ b/resources/macos_user.rb
@@ -14,7 +14,7 @@ action_class do
   end
 
   def setup_assistant_plist
-    ::File.join(user_home, '/Library/Preferences/com.apple.SetupAssistant.plist')
+    ::File.join(user_home, 'Library', 'Preferences', 'com.apple.SetupAssistant.plist')
   end
 
   def setup_assistant_keypair_values
--- a/resources/macos_user.rb
+++ b/resources/macos_user.rb
@@ -10,7 +10,7 @@ property :groups, [Array, String]
 
 action_class do
   def user_home
-    ::File.join('/Users', new_resource.username)
+    ::File.join('/', 'Users', new_resource.username)
   end
 
   def setup_assistant_plist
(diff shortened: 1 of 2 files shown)


americanhanko:
Excellent! Thanks for figuring this out.

test/integration/default/controls/users_and_groups.rb
```diff
@@ -18,9 +18,23 @@
     its('stdout.split') { should include '80' }
   end
 
-  describe command("/usr/libexec/PlistBuddy -c 'Print :autoLoginUser' /Library/Preferences/com.apple.loginwindow.plist") do
-    its('exit_status') { should eq 0 }
-    its('stdout.chomp') { should eq 'randall' }
+  setup_assistant_keypair_values = { 'autoLoginUser' => 'randall',
+                                     'lastUser' => 'loggedIn',
+                                   }
```
brownmike: Nit: Formatting

IMO
```ruby
foo = {
  'bar' => 'baz',
  'qux' => 'cddddddr',
}
```

resources/macos_user.rb
```diff
@@ -10,17 +10,30 @@
...
+    ::File.join('/', 'Users', new_resource.username)
   end
 
   def setup_assistant_plist
-    ::File.join(user_home, '/Library/Preferences/com.apple.SetupAssistant.plist')
+    ::File.join(user_home, 'Library', 'Preferences', 'com.apple.SetupAssistant.plist')
+  end
+
+  def login_window_plist
+    ::File.join('/', 'Library', 'Preferences', 'com.apple.loginwindow.plist')
```
brownmike: Just return a raw string IMO `'/Library/Preferences/com.apple.loginwindow.plist'`

Issue #107, opened 2018-05-02
macos_user resource fails to delete users


mjmerin:
Using the `macos_user` resource to delete an existing user does not work. 

As per the `macos_user` [readme](https://github.com/Microsoft/macos-cookbook/blob/master/documentation/resource_macos_user.md)

     :delete

      Delete a user specified by the macos_user username property.

I try to delete the previously created johnny account by doing the following

```ruby
macos_user 'delete johnny' do
   username 'johnny'
   action :delete
end
```

and unfortunately get

```
* macos_user[delete johnny] action delete

           ================================================================================
           Error executing action `delete` on resource 'macos_user[delete johnny]'
           ================================================================================

           Chef::Exceptions::ValidationFailed
           ----------------------------------
           name is a required property
```

There are also no integration tests that test for user deletion either.

PR #117, opened 2018-06-25
Fix for Issue #107: User Deletion

mjmerin:
## Bug Fix
The bug lies in the following: 

    execute "delete user: #{user}" do

`user` does not exist, so it was replaced with `new_resource.username`. 

### Created a user deletion recipe and test suite
Recipe creates a user in an execute block with `sysadminctl` to avoid using macos_user (in case there are future issues in it), and also to simulate a previously existing user.

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -115,6 +115,13 @@ suites:
     - admin-user
     - standard-user
 
+- name: delete_users
+  run_list:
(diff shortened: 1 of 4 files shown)


jazaval:
Thanks!

jazaval:
Thanks!

PR #119, opened 2018-06-26
update base boxes

jazaval:
This PR updates the base boxes used to the new standard, and resolves a bug where disk formatting causes the VM instance to hang.


tas50:
Any chance these will ever become public?

jazaval:
We definitely can't distribute the boxes themselves (Apple EULA and all), however @americanhanko is working on bringing our box creation tool up to a point where we could release _that_ and have people use their own macOS installers.

tas50:
Good ol EULAs.

PR #120, opened 2018-06-27
Use password to unlock keychain

MrChoclate:


jazaval:
Thanks for this! We should be releasing 2.3 to the Supermarket later today.

PR #121, opened 2018-06-28
Release/2.3

jazaval:
This releases adds a `hidden` property to the `macos_user` resource, and fixes a bug where deletion of a user was failing (thanks @mjmerin!). 

It also resolves a bug in the `certificate` resource for non-Vagrant use cases (thanks @MrChoclate!).

Finally, we've moved to a new set of internal Vagrant macOS boxes, which have a much more minimal initial configuration (thanks @americanhanko!), ensuring more of our resources work from an out-of-box macOS experience.

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -113,6 +113,7 @@ suites:
     controls:
     - admin-user
     - standard-user
+    - hidden-user
 
 - name: delete_users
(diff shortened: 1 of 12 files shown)


mjmerin:
We gotta bump that version number to 2.3!

Issue #123, opened 2018-07-18
Remove Deprecated Feature: homebrew_cask and homebrew_tap


mjmerin:
Most recent build of the `macos_cookbook` reveals the following

```
       Deprecated features used!
         Resource homebrew_cask has been loaded from a cookbook. The resource homebrew_cask is now included in Chef and will take precedence over the existing cookbook resource in the next major release of Chef (15.0, April 2019). You may be able to remove this cookbook dependency from your runlist if you do not use other recipes/resources/libraries from the cookbook. Alternatively there may be a newer version of this cookbook without the homebrew_cask resource. at 1 location:
           - /opt/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.3.37/lib/chef/log.rb:51:in `caller_location'
          See https://docs.chef.io/deprecations_map_collision.html for further details.
         Resource homebrew_tap has been loaded from a cookbook. The resource homebrew_tap is now included in Chef and will take precedence over the existing cookbook resource in the next major release of Chef (15.0, April 2019). You may be able to remove this cookbook dependency from your runlist if you do not use other recipes/resources/libraries from the cookbook. Alternatively there may be a newer version of this cookbook without the homebrew_tap resource. at 1 location:
           - /opt/chef/embedded/lib/ruby/gems/2.5.0/gems/chef-14.3.37/lib/chef/log.rb:51:in `caller_location'
          See https://docs.chef.io/deprecations_map_collision.html for further details.
```

mjmerin:
Resolved in Release 2.4

PR #125, opened 2018-08-02
Remove depends

mjmerin:
Fix for issue #123

Removing the dependency on homebrew from metadata removes the warnings during builds.


tas50:
This makes it a Chef 14+ cookbook only. You might wanna hold off on this

ehanlon:
Good point @tas50. We should at least target this for v3 instead.

jazaval:
Do we actually depend on the homebrew cookbook anymore?

Issue #45, opened 2018-02-12
Spotlight resource does not verify metadata server is running 


jazaval:
Spotlight resource assumes that the metadata server is running and hasn't been disabled with `launchctl unload`.

This requires a workaround of adding a `launchd` resource to start it and a small delay block to allow for it to finish spinning up before the `spotlight` resource is called:

```
launchd 'metadata server' do 
  path '/System/Library/LaunchDaemons/com.apple.metadata.mds.plist' 
  action :enable 
end 
 
ruby_block 'wait for metadata server spin-up' do 
  block { sleep 2 } 
end

spotlight '/' 
```

mjmerin:
Resolved in Release 2.4

PR #126, opened 2018-08-14
Fix for Issue #45: Spotlight resource does not verify metadata server is running

mjmerin:
- Support an additional status_flag for `mdutil_possible_states`
- Primarily focused in putting additional logic in the `metadata_util` library which now allows checking for server up/down status and for toggling. 
- Add unit test to check for new status flag
- Update integration test to actually turn off the Spotlight Server at the very start of the recipe

### DIFF
--- a/libraries/metadata_util.rb
+++ b/libraries/metadata_util.rb
@@ -3,23 +3,21 @@ include Chef::Mixin::ShellOut
 module MacOS
   class MetadataUtil
     attr_reader :status_flags
+    attr_reader :mdutil_output
 
     def initialize(volume)
       mdutil_possible_states = { 'Indexing enabled.' => ['on', ''],
                                  'Indexing disabled.' => ['off', ''],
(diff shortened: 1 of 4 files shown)


libraries/metadata_util.rb
```diff
@@ -7,14 +7,29 @@ class MetadataUtil
     def initialize(volume)
       mdutil_possible_states = { 'Indexing enabled.' => ['on', ''],
                                  'Indexing disabled.' => ['off', ''],
-                                 'Indexing and searching disabled.' => ['off', '-d'] }
+                                 'Indexing and searching disabled.' => ['off', '-d'],
+                                 'Spotlight server is disabled.' => %w(dis dis) }
```
jazaval: I'm not sure that this belongs in the states table, because it's kind of an exceptional case. I think it would be clearer if `server_disabled?` just parsed `mdutil_output` to determine if this case is true.

mjmerin: Changes have been added to the next set of commits.

Issue #122, opened 2018-07-10
Add CHANGELOG.md file to repo


Stromweld:
Add CHANGELOG.md file to repo with release notes for different versions. Makes it easier to see changes in chef supermarket as well as in the repo here without having to go through commits.

americanhanko:
Thanks for this issue, @Stromweld. We'll work this into our backlog.

mjmerin:
Resolved in Release 2.4

PR #130, opened 2018-08-15
Release/2.4

jazaval:
### Added
- Added `CHANGELOG.md`. ([Issue #122](https://github.com/Microsoft/macos-cookbook/issues/122)).
 ### Removed
- `homebrew` cookbook dependency removed. `homebrew_cask` and `homebrew_tap` is being deprecated by Chef and has not been used by `macos-cookbook` since version `2.0`. ([Issue #123](https://github.com/Microsoft/macos-cookbook/issues/123)).
 ### Fixed
- Fixed `keychain` resource documentation link.
- Update `metadata_util` library to consider Spotlight server status before manipulating indexing state. ([Issue #45](https://github.com/Microsoft/macos-cookbook/issues/45)).

### DIFF
--- a/resources/spotlight.rb
+++ b/resources/spotlight.rb
@@ -32,8 +32,15 @@ action_class do
 end
 
 action :set do
+  volume = MetadataUtil.new(target_volume)
+
+  execute 'Enable Spotlight server' do
+    command volume.toggle_spotlight_server(true)
+    only_if { volume.server_disabled? }
+  end
+
   execute "turn Spotlight indexing #{state} for #{target_volume}" do
     command mdutil + desired_spotlight_state.insert(0, '-i')
-    not_if { MetadataUtil.new(target_volume).status_flags == desired_spotlight_state }
+    not_if { volume.status_flags == desired_spotlight_state }
   end
 end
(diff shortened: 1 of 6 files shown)


test/cookbooks/macos_test/recipes/spotlight.rb
```diff
@@ -11,6 +11,10 @@
   end
 end
 
+execute 'test' do
+  command ['sudo', 'launchctl', 'unload', '-w', '/System/Library/LaunchDaemons/com.apple.metadata.mds.plist']
+end
+
```
americanhanko: This would be a good place to functionally test the built-in `macosx_service` resource, too. See [this example](https://github.com/americanhanko/vsts-agent-macos-cookbook/blob/2cefa59c48eb4baf410bee7a741ec9d65c49a551/recipes/bootstrap.rb#L114-L118) from `vsts_agent_macos`. The `plist` key is undocumented.

2. I could be missing something, but is there any particular `sudo` used here? Chef runs as `root` by default unless we use the `user` property.

resources/spotlight.rb
```diff
@@ -32,8 +32,15 @@ def desired_spotlight_state
 end
 
 action :set do
+  volume = MetadataUtil.new(target_volume)
+
+  execute 'Enable Spotlight server' do
+    command volume.toggle_spotlight_server(true)
```
americanhanko: I'm pretty sure we can use the `macosx_service` resource here. See my other comment.

PR #133, opened 2018-08-16
Use macosx_service for spotlight server on/off

mjmerin:
Responding to @americanhanko  review point, the use of the macosx_service resource guarantees idempotentiality and has allowed the removal of routines that directly call the launchctl command.

### DIFF
--- a/resources/spotlight.rb
+++ b/resources/spotlight.rb
@@ -34,7 +34,7 @@ end
 action :set do
   volume = MetadataUtil.new(target_volume)
 
-  macosx_service 'spotlight server' do
+  service 'spotlight server' do
     service_name 'mds'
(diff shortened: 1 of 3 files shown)


resources/spotlight.rb
```diff
@@ -34,9 +34,10 @@ def desired_spotlight_state
 action :set do
   volume = MetadataUtil.new(target_volume)
 
-  execute 'Enable Spotlight server' do
-    command volume.toggle_spotlight_server(true)
-    only_if { volume.server_disabled? }
+  macosx_service 'spotlight server' do
```
tas50: you should be able to just make this service. Chef does the right thing and wires up the provider under the hood

PR #135, opened 2018-08-29
Add an attribute to manage VNC screen lock setting

americanhanko:
## This PR

Adds a new attribute to manage the VNC screen lock setting

## Why

When using Apple Remote Desktop to control computers, disconnecting from a machine will automatically bring it into a 'sleep' state - this causes the screen to lock automatically given the screen lock setting in System Preferences -> Security & Privacy, which is sometimes a hindrance when needing to observe the state of the machine regularly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,11 +58,14 @@ to always keep macOS on and available.
 
 **Usage:** `include_recipe 'macos::keep_awake'`
 
-| Attributes used                         | Default value           |
-|-----------------------------------------|-------------------------|
-| `node['macos']['remote_login_enabled']` | `true`                  |
-| `node['macos']['network_time_server']`  | `'time.windows.com'`    |
-| `node['macos']['time_zone']`            | `'America/Los_Angeles'` |
+| Attributes used                            | Default value           |
+|--------------------------------------------|-------------------------|
+| `node['macos']['remote_login_enabled']`    | `true`                  |
+| `node['macos']['disable_vnc_screen_lock']` | `false`                 |
+| `node['macos']['network_time_server']`     | `'time.windows.com'`    |
+| `node['macos']['time_zone']`               | `'America/Los_Angeles'` |
+
+:warning: Disabling VNC screen lock could be considered a security risk at some organizations. Use with caution.
 
 ### Xcode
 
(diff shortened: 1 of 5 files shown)


attributes/default.rb
```diff
@@ -4,6 +4,7 @@
 default['macos']['xcode']['version'] = '9.4.1'
 
 default['macos']['remote_login_enabled'] = true
+default['macos']['disable_vnc_screen_lock'] = false
```
jazaval: Double negative here - maybe `‘screen_lock_enabled’ = true`?

AndreShields: Is the vnc screen lock separate and in a different plist than the default desktop screen lock? If so, would we not want to still include vnc? `‘vnc_screen_lock_enabled’ = true`

recipes/keep_awake.rb
```diff
@@ -73,5 +73,5 @@
 
 defaults 'disable VNC screen lock' do
   domain '/Library/Preferences/com.apple.RemoteManagement'
-  settings 'RestoreMachineState' => node['macos']['disable_vnc_screen_lock']
+  settings 'RestoreMachineState' => node['macos']['vnc_screen_lock_enabled']
```
jazaval: You'll need `!node['macos']['vnc_screen_lock_enabled']` here.

README.md
```diff
@@ -58,11 +58,14 @@ to always keep macOS on and available.
...
-|-----------------------------------------|-------------------------|
-| `node['macos']['remote_login_enabled']` | `true`                  |
-| `node['macos']['network_time_server']`  | `'time.windows.com'`    |
-| `node['macos']['time_zone']`            | `'America/Los_Angeles'` |
+| Attributes used                            | Default value           |
+|--------------------------------------------|-------------------------|
+| `node['macos']['remote_login_enabled']`    | `true`                  |
+| `node['macos']['vnc_screen_lock_enabled']` | `true`                  |
+| `node['macos']['network_time_server']`     | `'time.windows.com'`    |
+| `node['macos']['time_zone']`               | `'America/Los_Angeles'` |
```
brownmike: It's probably easier to specify the time zone in terms of a UTC offset so people don't have to guess at what magical string changes their time zone. That or list all the possible magic strings in the documentation.

Also, where does this get used?

brownmike: Nvm, I just realized this is a change to formatting to add a new attribute.

jazaval: we should probably map UTC offsets to the magic strings anyway - thanks for the suggestion!

Issue #129, opened 2018-08-15
Broken links in resource documentation


mjmerin:
Apple has changed and/or deprecated links the `macos-cookbook` is using. 

In [certificates](https://github.com/Microsoft/macos-cookbook/blob/master/documentation/resource_certificate.md), the `learn more about security` link is broken.

In [keychain](https://github.com/Microsoft/macos-cookbook/blob/master/documentation/resource_keychain.md), the `learn more about security` link is broken.

In [plist](https://github.com/Microsoft/macos-cookbook/blob/master/documentation/resource_plist.md), the links for `defaults`, `plutil`, and `PlistBuddy` are all broken.

In [spotlight](https://github.com/Microsoft/macos-cookbook/blob/master/documentation/resource_spotlight.md), the `learn more about spotlight` link is broken.

americanhanko:
As far as I can tell, Apple removed the online man pages.

PR #149, opened 2018-10-04
 Issue#129 Resolving broken links in README

felipedf:
Addressing to: https://github.com/Microsoft/macos-cookbook/issues/129

Well, seems like apple really removed online man pages.
This was a workaround to keep the info on the README, otherwise I guess we should just remove the links. For the `PlistBuddy` link tho, I linked that to a totally different man page, please look into that.


jazaval:
Hey @felipedf - thanks for your contribution. As it turns out, the team would prefer to remove these links if they're not being regularly updated by Apple. 

We're going to end up referencing the man pages, but not have direct links in the future.

PR #155, opened 2018-10-24
Add multi-converge testing for all suites, idempotency for select suites

jazaval:
This PR adds multi-converge testing for all suites, and idempotency enforcement for `spotlight` and `keep_awake` testing. 

It also fixes bugs related to enabling these options in `spotlight` and `keep_awake`.

### DIFF
--- a/libraries/metadata_util.rb
+++ b/libraries/metadata_util.rb
@@ -8,7 +8,8 @@ module MacOS
     def initialize(volume)
       mdutil_possible_states = { 'Indexing enabled.' => ['on', ''],
                                  'Indexing disabled.' => ['off', ''],
-                                 'Indexing and searching disabled.' => ['off', '-d'] }
+                                 'Indexing and searching disabled.' => ['off', '-d'],
+                                 'Error' => ['', ''] }
 
       @mdutil_output = shell_out('/usr/bin/mdutil', '-s', volume).stdout
       @status_flags = unless server_disabled?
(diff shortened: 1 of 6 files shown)


resources/system_preference.rb
```diff
@@ -20,11 +20,8 @@
     converge_by "set #{new_resource.preference} to #{new_resource.setting}" do
       set_pref = ['-set', new_resource.preference.to_s].join('')
       execute ['/usr/sbin/systemsetup', set_pref, new_resource.setting]
-      ruby_block 'sleep one second' do
-        block do
-          sleep 1
-        end
-      end
+      execute ['/usr/sbin/systemsetup', set_pref, new_resource.setting]
```
americanhanko: Not crazy about this. If we see see through CI that it resolved the issue, I'd like to implement better design.

libraries/metadata_util.rb
```diff
@@ -8,16 +8,11 @@ class MetadataUtil
     def initialize(volume)
       mdutil_possible_states = { 'Indexing enabled.' => ['on', ''],
                                  'Indexing disabled.' => ['off', ''],
-                                 'Indexing and searching disabled.' => ['off', '-d'] }
+                                 'Indexing and searching disabled.' => ['off', '-d'],
+                                 'Error' => ['', ''] }
```
americanhanko: UT for this?

PR #158, opened 2018-10-25
Feature/fix doc links

v-anshie:
This PR removes the links to outdated apple documentation in each resource documentation, and adds links to each resource that is being defined in the documentation.

### DIFF
--- a/documentation/resource_ard.md
+++ b/documentation/resource_ard.md
@@ -1,8 +1,8 @@
 ard
 ===
 
-Use the [**ard**](https://github.com/Microsoft/macos-cookbook/blob/master/resources/ard.rb) resource to manage the "Remote Management" settings, found in System
-Preferences > Sharing > Remote Management. Under the hood, the **ard** resource
+Use the **ard** resource to manage the "Remote Management" settings, found in System
+Preferences > Sharing > Remote Management. Under the hood, the [**ard**](https://github.com/Microsoft/macos-cookbook/blob/master/resources/ard.rb) resource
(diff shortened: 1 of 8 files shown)


documentation/resource_ard.md
```diff
@@ -23,15 +20,15 @@ end
...
+- `:configure` configures the agent using the `kickstart` default commandline arguments.
 
 The default `:configure` action is equivalent to the following
 **System Preferences > Sharing** settings:
 
 ![Sharing Preferences](sharing_preferences.png)
 
 The full syntax for all of the properties that are available to the **ard**
-resource is:
+resource are:
```
jazaval: This should be `is` since we're referring to the singular `syntax`, not the plural `properties`.

v-anshie: You are listing all of the properties though. So you are also referring to the plural properties?

documentation/resource_ard.md
```diff
@@ -1,17 +1,14 @@
 ard
 ===
 
-Use the **ard** resource to manage the "Remote Management" settings, found in System
-Preferences > Sharing > Remote Management. Under the hood, an **ard** resource
+Use the [**ard**](https://github.com/Microsoft/macos-cookbook/blob/master/resources/ard.rb) resource to manage the "Remote Management" settings, found in System
```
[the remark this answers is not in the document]
v-anshie: Makes sense, I will do this for all of the resources that I made this change too. Thanks!

documentation/resource_ard.md
v-anshie: Changed these back, thanks for the clarification.

PR #162, opened 2018-11-05
 Add dependency on chef-sugar cookbook to provide platform helper methods

ehanlon:
This allows us to utilize sugar like:
- `mac_os_x?`
- `virtual?`
- `mac_os_x_before_or_at_maverick?`

### DIFF
--- a/metadata.rb
+++ b/metadata.rb
@@ -12,4 +12,4 @@ issues_url 'https://github.com/Microsoft/macos-cookbook/issues'
 
 supports 'mac_os_x'
 
-depends 'chef-sugar'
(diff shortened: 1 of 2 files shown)


metadata.rb
```diff
@@ -11,3 +11,5 @@
 issues_url 'https://github.com/Microsoft/macos-cookbook/issues'
 
 supports 'mac_os_x'
+
+depends 'chef-sugar'
```
mjmerin: Suggest pinning. 

```suggestion
depends 'chef-sugar', '~> 4.1' 
```

PR #167, opened 2018-11-15
Release/2.8

americanhanko:


v-anshie:
Merging since the build failure is due to the nodes not being able to install xCode in the allotted time. Verified locally.

PR #169, opened 2018-11-29
Remove hard-coded path for xcode-install gem

### LINKED ISSUE #168: XCVersion is hard coded
The helper for xcversion fails if Chef is installed in another place, this is common if you're running using chefdk/chef-workstation. 

https://github.com/Microsoft/macos-cookbook/blob/74177c45c368c36bb4a2b44db1ace025baea5945/libraries/xcversion.rb#L7

jazaval:
### Fixed
- Fixed an issue where the path for the `xcversion` utility was hard-coded when installed as a Chef gem, which caused failures when converging with ChefDK or Workstation.

Resolves #168.

### DIFF
--- a/libraries/xcversion.rb
+++ b/libraries/xcversion.rb
@@ -3,8 +3,8 @@ include Chef::Mixin::ShellOut
 module MacOS
   module XCVersion
     class << self
-      def xcversion
-        '/opt/chef/embedded/bin/xcversion '.freeze
+      def xcversion_path
+        Chef::Util::PathHelper.join(Chef::Config.embedded_dir, 'bin', 'xcversion')
       end
 
       def update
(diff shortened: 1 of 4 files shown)


libraries/xcversion.rb
```diff
@@ -3,37 +3,45 @@
...
   module XCVersion
     class << self
-      def xcversion
-        '/opt/chef/embedded/bin/xcversion '.freeze
+      def xcversion_path
+        Chef::Util::PathHelper.join(Chef::Config.embedded_dir, 'bin', 'xcversion')
+      end
+
+      def xcversion(command)
+        [xcversion_path, command].join(' ')
```
americanhanko: You could also write this function as:


```ruby
def xcversion(subcommand, *args)
  [xcversion, subcommand, *args].join ' '
end
```

This would allow us to specify each subcommand and subsequent arguments separately, rather than string interpolating all the arguments including the subcommand.

PR #173, opened 2018-12-28
Added to and Moved plist tests

chrismarlin:
per Marks recommendation, moved plist tests to a separate file.


v-anshie:
Does not build or pass any tests

PR #175, opened 2019-01-07
Bump chef-sugar version

ehanlon:
This is for a downstream dependency in one of our internal cookbooks.


jazaval:
Looks like the breaking change to `5.0` for `chef-sugar` is going to Chef 13+, which we're already at or above. Should be safe as a patch release.

Issue #174, opened 2019-01-07
[FEATURE] Expose `xcversion install --url`


ehanlon:
## Describe the feature  
Expose the `--url` flag for `xcversion install` as a property of the `xcode` resource.

## Describe the reasoning behind the feature  
This allows installation of Xcode from a custom file path or URL (e.g. CDN, storage account) reducing the dependency on [developer.apple.com](https://developer.apple.com), and removing the need for setting Apple ID secrets.

PR #179, opened 2019-01-16
Release/2.10.0

ehanlon:
- Add `command_line_tools` resource
- Update default Xcode version to 10.1 for `macos::xcode` recipe
- `correct-horse-battery-staplesudo`
- New Feature Issue #174: Expose --url option for xcversion install


spec/unit/libraries/xcode_spec.rb
```diff
@@ -59,28 +59,50 @@
                    )
     end
     it 'returns the name of Xcode 10 GM when initialized with the semantic version' do
-      xcode = MacOS::Xcode.new('10.0', '/Applications/Xcode.app')
+      xcode = MacOS::Xcode.new('10.0', '/Applications/Xcode.app', '')
```
americanhanko: Needing to actually pass the empty string doesn't feel like a great design, it's not clear why the Xcode class is being initialized with this value. Could we set the default value to an empty string in the initializer instead of `nil`? Alternatively, we could make the `download_url` parameter a keyword argument.

README.md
```diff
@@ -1,5 +1,4 @@
-macOS Cookbook
-==============
+# macOS Cookbook
```
americanhanko: Why was the markdown changed here? I think the syntax that was there previously makes the raw file more readable, and it's the same functionally.

documentation/resource_command_line_tools.md
```diff
@@ -0,0 +1,49 @@
...
+
+The ``command_line_tools`` resource has the following actions:
+
+``:install``
+
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Default. Install the latest Xcode Command Line Tools.
+
+``:nothing``
+
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;This resource block does not act unless notified by another resource to take action. Once notified, this resource block either runs immediately or is queued up to run at the end of the Chef Client run.
```
americanhanko:
```suggestion
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;This resource block does not act unless notified by another resource to take action. Once notified, this resource block either runs immediately or is queued up to run at the end of the `chef-client` run.
```

documentation/resource_command_line_tools.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Examples
+
+### Install Xcode Command Line Tools
+
+```ruby
+command_line_tools 'random_name'
+```
+
+If running **macOS 10.14.2**, it will install **'Command Line Tools (macOS Mojave version 10.14) for Xcode-10.14'** via the Software Update Tool.
```
americanhanko:
```suggestion
If running **macOS 10.14.2**, it will install **'Command Line Tools (macOS Mojave version 10.14) for Xcode-10.14'** via the `softwareupdate` utility.
```

libraries/xcode.rb
```diff
@@ -6,12 +6,18 @@ class Xcode
...
 
-    def initialize(semantic_version, intended_path)
+    def initialize(semantic_version, intended_path, download_url)
       @semantic_version = semantic_version
       @intended_path = intended_path
       @apple_version = Xcode::Version.new(@semantic_version).apple
-      @version = XCVersion.available_versions[xcode_index(@apple_version)].strip
+      @download_url = download_url
+      @version = if download_url.empty?
+                   XCVersion.available_versions[xcode_index(@apple_version)].strip
```
americanhanko: Let's break this up to make it a little more readable, and avoid some potentially cryptic errors due to the method chaining.

```suggestion
                   version = xcode_index @apple_version
                   listed_version = XCVersion.available_versions[version]
                   listed_version.strip
```

spec/unit/libraries/xcode_spec.rb
jazaval: Sure.

README.md
ehanlon: These changes were from `markdownlint`

PR #180, opened 2019-01-16
New Xcode Property: download_url

mjmerin:
## Describe the Pull Request  
expose the`--url` option of xcversion to specify an alternate download url for Xcode. This fixes Issue #174 

## Justification  
As mentioned in the Issue #174, This allows installation of Xcode from a custom file path or URL (e.g. CDN, storage account) reducing the dependency on developer.apple.com, and removing the need for setting Apple ID secrets.


libraries/xcversion.rb
```diff
@@ -31,8 +31,18 @@ def list_available_xcodes
         xcversion 'list'
       end
 
+      def xcode_install_options(xcode)
+        options = ''
+
+        unless xcode.download_url.empty?
+          options = "--url '#{xcode.download_url}'"
```
jazaval: This seems too specific for the method name - I know this is the only option we're using at the moment, but that makes me think the method name should actually be `download_url_option` or something.

libraries/xcversion.rb
```diff
@@ -31,8 +31,18 @@ def list_available_xcodes
         xcversion 'list'
       end
 
+      def xcode_install_options(xcode)
```
jazaval:
```suggestion
      def download_url_option(xcode)
```

libraries/xcversion.rb
```diff
@@ -31,8 +31,18 @@ def list_available_xcodes
...
+        unless xcode.download_url.empty?
+          options = "--url '#{xcode.download_url}'"
+        end
+
+        options
+      end
+
       def install_xcode(xcode)
-        xcversion "install '#{xcode.version}'"
+        xcversion "install #{xcode.version} #{xcode_install_options(xcode)}"
```
jazaval:
```suggestion
        xcversion "install #{xcode.version} #{download_url_option(xcode)}"
```

libraries/xcversion.rb
mjmerin: good idea!

PR #191, opened 2019-02-25
Replace ard resource with remote_management

jazaval:
## This PR
Replaces the `ard` resource with `remote_management` and updates applicable tests and documentation.

The new `remote_management` resource greatly simplifies syntax and reduces the needed macOS domain knowledge around `kickstart` options. However, it has less functionality than `ard` and is a significant breaking change. 

## Why
The `ard` resource is only idempotent with default properties and actions; `remote_management` is fully idempotent. From our known use cases, most of the additional configuration options are not used. 

Additionally, making the `ard` resource idempotent would require a library capable of reversing the permissions bitmask set in the `ARD_AllLocalUsersPrivs` key. We'll consider adding this functionality to `remote_management` if there's some demand for more granular configuration. 

Fixes AB#2576657


resources/remote_management.rb
```diff
@@ -0,0 +1,24 @@
...
+default_action :enable
+
+kickstart = '/System/Library/CoreServices/RemoteManagement/ARDAgent.app/Contents/Resources/kickstart'
+
+action :enable do
+  execute "#{kickstart} -activate" do
+    not_if { RemoteManagement.activated? }
+  end
+
+  execute "#{kickstart} -configure -allowAccessFor -allUsers -access -on -privs -all" do
```
ehanlon: Are we eventually going to have the ability to specify allowed users?

libraries/remote_management.rb
```diff
@@ -0,0 +1,40 @@
...
+        ::File.exist?('/Library/Application Support/Apple/Remote Desktop/RemoteManagement.launchd')
+      end
+
+      def configured?
+        RemoteManagement.plist.include?('ARD_AllLocalUsers = true') &&
+          RemoteManagement.plist.include?('ARD_AllLocalUsersPrivs = 1073742079') &&
+          RemoteManagement.plist.include?(full_privileges)
+      end
+
+      def plist
```
americanhanko: Could we name this method `plist_contents`? Current name sounds like it's returning the location of the plist file itself, when in reality, it's returning the plist contents. It would make the implementation of the `configured?` a bit clearer, also.

libraries/remote_management.rb
```diff
@@ -0,0 +1,40 @@
+include Chef::Mixin::ShellOut
+
+module MacOS
+  class RemoteManagement
+    class << self
+      def activated?
+        ::File.exist?('/Library/Application Support/Apple/Remote Desktop/RemoteManagement.launchd')
+      end
+
+      def configured?
```
americanhanko: Perhaps a bit more explicit: `configured_with_full_privileges?`

libraries/remote_management.rb
```diff
@@ -0,0 +1,40 @@
+include Chef::Mixin::ShellOut
+
+module MacOS
+  class RemoteManagement
+    class << self
+      def activated?
+        ::File.exist?('/Library/Application Support/Apple/Remote Desktop/RemoteManagement.launchd')
```
americanhanko: Nit: the parentheses.

resources/remote_management.rb
```diff
@@ -0,0 +1,24 @@
+resource_name :remote_management
+default_action :enable
+
+kickstart = '/System/Library/CoreServices/RemoteManagement/ARDAgent.app/Contents/Resources/kickstart'
+
+action :enable do
+  execute "#{kickstart} -activate" do
```
americanhanko: Any particular reason we chose to use guards here instead of `load_current_value`?

spec/unit/resources/remote_management_spec.rb
```diff
@@ -0,0 +1,102 @@
+require 'spec_helper'
+
```
americanhanko: These unit tests look great.

resources/remote_management.rb
jazaval: There aren't any properties to `load_current_value` of.

jazaval: Are actions actually just properties?

jazaval: Is something like this possible?
```
load_current_value do
    enable RemoteManagement.activated?
end

action :enable do
  converge_if_changed :enable do
    execute "#{kickstart} -activate"
  end
end

PR #192, opened 2019-02-26
update azure-pipelines.yml test suites list

v-anshie:
---

name: Pull Request
about: Update azure-pipelines.yml
title: "Update azure-pipelines.yml"

---

## Describe the Pull Request  
- This PR adds test suites from the kitchen.yml to azure-piplines.yml that were previously not on the list. This allows the builds to test against all suites.

- It also gets suites passing that weren't being run previously and turned out to be failing.


kitchen.yml
```diff
@@ -54,7 +54,7 @@ suites:
   provisioner:
     enforce_idempotency: true
   run_list:
-  - recipe[macos_test::software_updates]
```
v-anshie: This change was made in the release branch already.

PR #200, opened 2019-06-03
Hotfix/xcode beta install

americanhanko:
# Install Xcode versions with  whitespace

## This PR
- Fixes an issue where Xcode versions containing whitespace were not properly quoted in command execution.
- Regression from version 2.10 release.

### DIFF
--- a/libraries/xcode.rb
+++ b/libraries/xcode.rb
@@ -43,7 +43,7 @@ module MacOS
 
     def current_path
       if installed_path.nil? && @version.match?(/beta/i)
-        "/Applications/Xcode-#{@apple_version}.Beta.app"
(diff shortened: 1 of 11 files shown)


libraries/xcode.rb
```diff
@@ -43,7 +43,7 @@ def installed_path
 
     def current_path
       if installed_path.nil? && @version.match?(/beta/i)
-        '/Applications/Xcode-beta.app'
+        "/Applications/Xcode-#{@apple_version}.Beta.app"
```
v-anshie: Where does this naming come from? I get the previous change when I download the beta from the developer site. Just curious is all.

PR #203, opened 2019-07-31
Fix Xcode beta installs and build pipeline

jazaval:
### Fixed
- Fixed an issue where Xcode versions containing whitespace were not properly quoted in command execution. Regression from version 2.10 release.

- Fixed an issue where the `box_version` was not properly specified in the `kitchen.yml`

Additional cookstyle updates littered throughout.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## [3.0.2] - 2019-06-03
+
+### Fixed
+- Fixed an issue where Xcode versions containing whitespace were not properly quoted in command execution. Regression from version 2.10 release.
+
 ## [3.0.1] - 2019-03-15
 
 Thanks to @jkronborg for these two fixes!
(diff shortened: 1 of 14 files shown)


libraries/system.rb
```diff
@@ -26,7 +26,7 @@ def initialize(virtualization_systems)
       end
 
       def vm?
-        @virtualization_systems.empty? || @virtualization_systems.values.include?('guest') ? true : false
+        @virtualization_systems.empty? || @virtualization_systems.value?('guest') ? true : false
```
ehanlon: I thought the conditional already returned a bool?

CHANGELOG.md
```diff
@@ -1,12 +1,18 @@
 # Changelog
 
+## [3.0.2] - 2019-06-03
```
ehanlon: 3.0.8?

libraries/system.rb
jazaval: They're your cookstyle changes ;-)

PR #209, opened 2019-09-25
Add path helper classes

jazaval:
## This PR
Adds a library of helper methods for reducing duplication in specifying macOS user and system paths. Thanks @mjmerin!

### DIFF
--- a/spec/unit/libraries/paths_spec.rb
+++ b/spec/unit/libraries/paths_spec.rb
@@ -1,48 +1,48 @@
 require 'spec_helper'
-include LabCore
+include MacOS
 
 user_name = 'test'
 
-describe LabCore::UserPath, 'UserPath library helper functions' do
+describe MacOS::UserPath, 'UserPath library helper functions' do
   context 'using UserPath library routines to create paths' do
     it 'home creates the correct path to user home' do
       expect(UserPath.home(user_name)).to eq "/Users/#{user_name}"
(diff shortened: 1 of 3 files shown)


spec/unit/libraries/paths_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+end
+
+describe MacOS::SystemPath, 'SystemPath library helper functions' do
+  context 'using SystemPath library routines to create paths' do
+    it 'directory creates the correct directory path' do
+      expect(SystemPath.directory('how', 'now', 'brown', 'cow')).to eq '/how/now/brown/cow'
+    end
+
+    it 'root creates the correct system path' do
+      expect(SystemPath.root).to eq '/'
```
ehanlon: Is this expected to change in Catalina?

chrismarlin:
________________________________

spec/unit/libraries/paths_spec.rb
chrismarlin: The ability to write to the system path you mean? Yeah I would expect it would fail, even as root.

jazaval: This is still valid in Catalina, but there are many folders at `/` that are no longer writeable. The OS handles the underlying "firmlinks" (Apple's term) to "Data".

PR #213, opened 2019-10-09
Release/3.1.0

jazaval:
### Added

- Added support for macOS Catalina
- Added support for Chef 15
- Added `UserPath` and `SystemPath` libraries to simplify calls to macOS filepaths

### Fixed

- Fixed an issue parsing new `softwareupdate` output on macOS Catalina
- Fixed a dependency issue with `xcode-install` on Chef 15
- Fixed and updated the `xcode` resource platform compatibilty check (thanks to https://xcodereleases.com for tracking this)

### Removed

- Removed testing on macOS Sierra
- Removed testing on Chef 14
- Removed dependency on `chef-sugar`


CHANGELOG.md
```diff
@@ -1,5 +1,25 @@
 # Changelog
 
+## [3.1.0] - 2019-10-09
+
+### Added
+
+- Added support for macOS Catalina
+- Added support for Chef 15
+- Added `UserPath` and `SystemPath` libraries to simplify calls to macOS filepaths
```
ehanlon: Does every line under the `Added` heading need to start with `Added`?

jazaval: I prefer full sentences grammatically, but after checking https://keepachangelog.com/en/1.0.0/, it seems you're correct.

libraries/command_line_tools.rb
```diff
@@ -1,34 +1,59 @@
...
+    end
+
+    def catalog_recommendation_label
+      if platform_specific.empty?
+        Chef::Log.info "No Command Line Tools specific to #{macos_version} available from Software Update Catalog, selecting by maximum Xcode version."
+        versions = available.map { |product| Gem::Version.new xcode_version(product) }
+        available.detect { |product| product.include? versions.max.version }
+      else
+        versions = platform_specific.map { |product| Gem::Version.new xcode_version(product) }
+        platform_specific.detect { |product| product.include? versions.max.version }
```
americanhanko: There's a little bit of duplication here – thinking we might be able to extract to a new method?

```ruby
def recommended_versions(software_update_package_list)
  versions = software_update_package_list.map { |product| Gem::Version.new xcode_version(product) }
  software_update_package_list.detect { |product| product.include? versions.max.version }
end
```

then:

```ruby
def catalog_recommendation_label
  unless platform_specific.empty?
    recommended_versions(available)
  else
    Chef::Log.info "No Command Line Tools specific to #{macos_version} available from Software Update Catalog"
    Chef::Log.info "Command Line Tools version will be selected using maximum Xcode version."
    recommended_versions(platform_specific)
  end
end
```

jazaval: Thanks - updated.

PR #217, opened 2019-12-07
Add upgrade action to command_line_tools resource

jazaval:
## This PR
- Adds an `:upgrade` action to the `command_line_tools` resource. This action will check for an updated version of Command Line Tools and install them if available.


chrismarlin:
With this addition are we ok with just the two existing contexts: 'with' and 'without libxcrun present' for the clt resource spec tests?

PR #219, opened 2020-03-04
Fix platform detection in system library

jazaval:
### Fixed

- Fixed an issue with the `system` library where a system without hypervisor 
software installed could be mistakenly identified as a VM and not have certain
platform-specific power preferences set.

### DIFF
--- a/libraries/system.rb
+++ b/libraries/system.rb
@@ -26,7 +26,7 @@ module MacOS
       end
 
       def vm?
-        @virtualization_systems.empty? || @virtualization_systems.value?('guest') ? true : false
+        @virtualization_systems.value?('guest') ? true : false
       end
     end
 
--- a/libraries/system.rb
+++ b/libraries/system.rb
@@ -26,7 +26,7 @@ module MacOS
       end
 
       def vm?
-        @virtualization_systems.value?('guest') ? true : false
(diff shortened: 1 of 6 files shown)


spec/unit/recipes/keep_awake_spec.rb
```diff
@@ -80,34 +80,6 @@
   end
 end
 
-shared_context 'running in an undetermined virtualization system' do
-  before(:each) do
-    chef_run.node.automatic['virtualization']['systems'] = {}
-    chef_run.node.automatic['hardware']['machine_model'] = ''
-  end
```
americanhanko:
```ruby
shared_context 'running in an undetermined virtualization system'
```
Or a bare metal without a hypervisor installed, apparently.

image

libraries/system.rb
```diff
@@ -26,7 +26,7 @@ def initialize(virtualization_systems)
       end
 
       def vm?
-        @virtualization_systems.empty? || @virtualization_systems.value?('guest') ? true : false
+        @virtualization_systems.value?('guest') ? true : false
```
americanhanko: Do we actually need the ternary?

```
def vm?
  @virtualization_systems.value?('guest')
end
```

jazaval: Good call.

PR #226, opened 2020-08-05
Release/3.4.0

### LINKED ISSUE #224: command_line_tools library doesn't support latest Big Sur betas
The regex in `command_line_tools` line 67 needs to be updated to support Big Sur now that `sw_vers` reports 11.0 as the `productVersion`, rather than 10.16.

https://github.com/microsoft/macos-cookbook/blob/6dc88804a15d2485b9ffb0903376f05dc7dd6fb8/libraries/command_line_tools.rb#L67

Suggested fix (tested locally):

`shell_out(['/usr/bin/sw_vers', '-productVersion']).stdout.chomp[/1[01]\.\d+/]`

I'm working on submitting my own pull request but having some issues with github auth.

jazaval:
# Changelog

### Added

- Support and testing for Big Sur. Due to a bug in the `kickstart` script in Big Sur, the `remote_management`
resource is not functional on any system that reports `11.0` instead of `10.16`, which is the case
on 20A5323l (beta 3) or later. This will be resolved at a later date. 

- Deprecation notice for `macos::keep_awake`.

### Fixed

- The `xcode` resource now installs the latest available beta revision or GM seed of Xcode if there 
is not an official release of the requested version. Note that installing the early "Xcode 12 for macOS 
Universal Apps" betas is not supported since Apple has already unified these into the core Xcode
bundles as of 12A8169g (beta 3).

- The `macos_user` resource now prevents the screen from re-locking at boot when auto-login is enabled.

- The `command_line_tools` OS version parsing regex has been fixed for macOS 11.0

- The `automatic_software_updates` resource tests have been fixed for macOS 11.0

Resolves #224.


libraries/command_line_tools.rb
```diff
@@ -64,7 +64,7 @@ def xcode_version(product)
     end
 
     def macos_version
-      shell_out(['/usr/bin/sw_vers', '-productVersion']).stdout.chomp[/10\.\d+/]
+      shell_out(['/usr/bin/sw_vers', '-productVersion']).stdout.chomp[/1[01]\.\d+/]
```
americanhanko, line 67: Arguably YAGNI, but it's worth considering implementing the suggested pattern for finding the semantic version from the output of this command. See [Is there a suggested regular expression (RegEx) to check a SemVer string?](https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string)

jazaval:
> @jazaval LGTM.
> 
> What's the plan for `kickstart` compatibility? Is this something that the cookbook will need to address or is Apple going to fix it upstream?

Thanks for the signoff. Should be an upstream fix, won't require any changes here other than unskipping the Big Sur `remote_management` resource tests.

libraries/command_line_tools.rb
jazaval: That's a good point - arguably this belongs in a more generic library like `MacOS::System` anyway. We'll make that move / improvement when we clean up that library in the `keep_awake` removal.

PR #229, opened 2021-02-10
Fix parsing of 'xcversion list' output

jazaval:
# Changelog

### Fixed

- Updated `xcversion` library `list` output to not include "(installed)" status, since this is determined separately in the `xcode` resource. The additional string content was breaking parsing that prevents installation of non-complete Xcode bundles such as `12 for macOS Universal Apps`.

- Fixes for Cookstyle version: 7.7.2

### Removed

- Removed ChefSpec testing on deprecated Fauxhai platforms

### DIFF
--- a/libraries/xcversion.rb
+++ b/libraries/xcversion.rb
@@ -53,6 +53,7 @@ module MacOS
       def available_versions
         lines = shell_out!(XCVersion.list_available_xcodes).stdout.lines
         lines.reject { |line| line.start_with?(/\D/) }
+        lines.map { |line| line.chomp(' (installed)') }
       end
     end
   end
--- a/libraries/xcversion.rb
+++ b/libraries/xcversion.rb
@@ -52,7 +52,7 @@ module MacOS
 
       def available_versions
         lines = shell_out!(XCVersion.list_available_xcodes).stdout.lines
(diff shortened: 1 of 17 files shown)


ehanlon (dismissed):

adambollen (dismissed):

libraries/xcversion.rb
```diff
@@ -53,10 +53,11 @@ def installed_xcodes
       def available_versions
         lines = shell_out!(XCVersion.list_available_xcodes).stdout.lines
         lines.reject { |line| line.start_with?(/\D/) }
+        lines.map { |line| line.chomp(' (installed)') }
```
GuMiner, line 56: Removing the extra space before `(installed)` seems a bit weird. Would trimming all leading/trailing whitespace make more sense?
```suggestion
        lines.map { |line| line.chomp('(installed)').strip }
```

jazaval: That's definitely more explicit and readable, it would be easy to miss the space in a refactor.

adambollen (dismissed):

GuMiner (approved):

adambollen (approved):

GuMiner (approved):

hushlorentz (approved):

PR #242, opened 2021-10-15
Fix defaults data conversion

GuMiner:
## Fixes
Defaults data conversion now handles spaces in strings, arrays, and dictionaries

## Justification  
Make defaults work!

## Details
The original data conversion code converted Ruby data types naively into strings. Unfortunately, this format isn't in the format that `defaults` expects. 

This change modifies the data conversion process to correctly output the expected data in a format acceptable to `defaults`, while also adding in additional tests to verify the output of the conversion process.


jazaval:
Re-opening PR to trigger pipeline checks, these previously got disconnected due to a service connection update to ADO.

PR #248, opened 2022-03-04
QOL Improvment: Remove dependency on attribute and use property instead for keychain password in certificate resource

jazaval:
## Describe the Pull Request
When using the certificate resource install a certificate to a locked keychain, the password to unlock is only settable via the node['macos']['admin_password'] attribute.

This means users of the certificate resource will have to re-declare this above any calls to the resource unless it's in kitchen.

This PR, remove that logic and instead introduces a kc_passwd property similar to the keychain resource to allow users to input this password.

## Justification
Not really a bug but a QOL improvement.

### DIFF
--- a/resources/certificate.rb
+++ b/resources/certificate.rb
@@ -6,15 +6,14 @@ default_action :install
 property :certfile, String
 property :cert_password, String, sensitive: true
 property :keychain, String, required: true
-property :kc_passwd, String, sensitive: true
+property :kc_passwd, String, required: true, sensitive: true
 property :apps, Array
 
 action :install do
   cert = SecurityCommand.new(new_resource.certfile, new_resource.keychain)
 
   execute 'unlock keychain' do
-    password = new_resource.property_is_set?(:kc_passwd) ? new_resource.kc_passwd : node['macos']['admin_password']
-    command Array(cert.unlock_keychain(password))
+    command Array(cert.unlock_keychain(new_resource.kc_passwd))
   end
 
   cert_shasum = shell_out("shasum #{new_resource.certfile}").stdout.upcase.gsub(/\s.+/, '')
--- a/resources/certificate.rb
+++ b/resources/certificate.rb
@@ -5,6 +5,7 @@ provides :certificate
 property :certfile, String
 property :cert_password, String
 property :keychain, String
+property :kc_passwd, String
 property :apps, Array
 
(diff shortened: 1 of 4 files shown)


mjmerin:
LGTM

mjmerin (approved):

v-anshie (approved):

resources/certificate.rb
```diff
@@ -1,26 +1,26 @@
...
-  cert = SecurityCommand.new(new_resource.certfile, keychain)
+  cert = SecurityCommand.new(new_resource.certfile, new_resource.keychain)
 
   execute 'unlock keychain' do
-    command Array(cert.unlock_keychain(node['macos']['admin_password']))
+    command Array(cert.unlock_keychain(new_resource.kc_passwd))
   end
 
+  cert_shasum = shell_out("shasum #{new_resource.certfile}").stdout.upcase.gsub(/\s.+/, '')
+  find_cert_output = shell_out("/usr/bin/security find-certificate -a -Z #{new_resource.keychain}").stdout
```
jweyer8, line 20: `/usr/bin/security find-certificate -a -Z | awk '/SHA/ {print $3}'` might optimize the `include?` a bit?

jweyer8 (approved):

resources/certificate.rb
```diff
@@ -1,26 +1,26 @@
...
 action :install do
-  cert = SecurityCommand.new(new_resource.certfile, keychain)
+  cert = SecurityCommand.new(new_resource.certfile, new_resource.keychain)
 
   execute 'unlock keychain' do
-    command Array(cert.unlock_keychain(node['macos']['admin_password']))
+    command Array(cert.unlock_keychain(new_resource.kc_passwd))
   end
 
+  cert_shasum = shell_out("shasum #{new_resource.certfile}").stdout.upcase.gsub(/\s.+/, '')
```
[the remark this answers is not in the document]
mjmerin: .strip would only remove the white space, what I'm doing here is also taking out anything after the whitespace.

resources/certificate.rb
mjmerin: I believe that finds the shasums of all certificates on all keychains. So depending on how many certs/keychains there are, this more direct approach is probably faster and easier to read. Again, I don't really have any metrics to confirm or deny.

PR #252, opened 2022-03-07
Add options for sensitivity where password is present. Fix docs for keychain kc_file

Brantone:
## Describe the Pull Request  
Minor modification on docs mentioning `kc_file` attribute rather than `keychain`
Add sensitivity options where passwords are applicable

## Justification  
Clear docs and password security


jazaval:
Thanks, @Brantone! I've updated the target branch to our upcoming 5.0 release and fixed the merge. Also, do you have any objections to requiring the appropriate resource calls to be `sensitive`, or at least making `sensitive` true by default?

jweyer8 (approved):

resources/certificate.rb
```diff
@@ -5,15 +5,16 @@
 
 property :certfile, String
 property :cert_password, String, sensitive: true
-property :keychain, String
+property :keychain, String, required: true
+property :kc_passwd, String, required: true, sensitive: true
```
Brantone, line 9: 👍 
Awesome, was thinking about doing similar on fork, but wasn't sure of use-case for `node['macos']['admin_password']`

Brantone:
>  do you have any objections to requiring the appropriate resource calls to be sensitive, or at least making sensitive true by default?

That would makes sense to me. I was thinking of doing similar, but wasn't sure about maintaining expected behaviour and communicating to other impending change.

jazaval:
> > do you have any objections to requiring the appropriate resource calls to be sensitive, or at least making sensitive true by default?
> 
> That would makes sense to me. I was thinking of doing similar, but wasn't sure about maintaining expected behaviour and communicating to other impending change.

We're bumping the major version of the cookbook, and assuming no one is consuming without a version pin for breaking changes (or is utilizing Policyfiles).

Brantone:
I might have further doc tweaks coming in ... very minor.

resources/certificate.rb
```diff
@@ -8,19 +8,22 @@
 property :keychain, String, required: true
 property :kc_passwd, String, required: true, sensitive: true
```
Brantone, line 9: I know already merged, just wondering about consistency in naming of "password" for `cert_password` vs. `cert_passwd` 🤔

Issue #234, opened 2021-05-11
[BUG] xcode resource attempts to auth with Apple even if the on-disk Xcode is up-to-date


aaronclarke:
## Describe the Bug 
The xcode resource attempts to auth with Apple even if the on-disk Xcode is up-to-date.
This is because the Xcode `installed?` method that we use to guard against re-install/upgrade shouldn't be tied up in that initialization, but it is.

## Error Message  
```
    ================================================================================
    Error executing action `install_xcode` on resource 'xcode[install Xcode 12.4]'
    ================================================================================
    
    Mixlib::ShellOut::ShellCommandFailed
    ------------------------------------
    Expected process to exit with [0], but received '1'
    ---- Begin output of /opt/chef/embedded/bin/xcversion update ----
    STDOUT: 
    STDERR: /opt/chef/embedded/lib/ruby/gems/2.7.0/gems/fastlane-2.182.0/spaceship/lib/spaceship/client.rb:909:in `handle_error': Access forbidden (Spaceship::AccessForbiddenError)
    	from /opt/chef/embedded/lib/ruby/gems/2.7.0/gems/fastlane-2.182.0/spaceship/lib/spaceship/client.rb:886:in `block in send_request'
    	from /opt/chef/embedded/lib/ruby/gems/2.7.0/gems/fastlane-2.182.0/spaceship/lib/spaceship/client.rb:646:in `with_retry'
    	from /opt/chef/embedded/lib/ruby/gems/2.7.0/gems/fastlane-2.182.0/spaceship/lib/spaceship/client.rb:882:in `send_request'
    	from /opt/chef/embedded/lib/ruby/gems/2.7.0/gems/fastlane-2.182.0/spaceship/lib/spaceship/client.rb:726:in `request'
    	from /opt/chef/embedded/lib/ruby/gems/2.7.0/gems/fastlane-2.182.0/spaceship/lib/spaceship/client.rb:481:in `send_shared_login_request'
    	from /opt/chef/embedded/lib/ruby/gems/2.7.0/gems/fastlane-2.182.0/spaceship/lib/spaceship/portal/portal_client.rb:28:in `send_login_request'
    	from /opt/chef/embedded/lib/ruby/gems/2.7.0/gems/fastlane-2.182.0/spaceship/lib/spaceship/client.rb:808:in `do_login'
[...]

PR #254, opened 2022-03-08
update `xcode` to not use attributes, only auth dev account if out-of-date

jazaval:
# Changelog

### Fixed

- Extracted authentication with Apple via `xcode` resource away `xcode` object instantiation, resolving #234. 

### Added

- Added `apple_id` property to `xcode` resource to remove dependency on attributes or data bags for authentication.

### Removed

- Removed the ability to authenticate with Apple for `xcode` downloads via node attributes or data bags.
- Removed last default cookbook attribute:`node['macos']['apple_id']`

### DIFF
--- a/libraries/xcode.rb
+++ b/libraries/xcode.rb
@@ -9,13 +9,14 @@ module MacOS
       @semantic_version = semantic_version
       @intended_path = intended_path
       @download_url = download_url
+      @version = -> { determine_version }.call
     end
 
     def determine_version
-      @version = if download_url.empty?
-                   latest_xcode_revision(Xcode::Version.new(semantic_version))
+      @version = if @download_url.empty?
+                   latest_xcode_revision(Xcode::Version.new(@semantic_version))
                  else
-                   semantic_version
+                   @semantic_version
                  end
     end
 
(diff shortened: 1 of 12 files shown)


libraries/xcode.rb
```diff
@@ -3,18 +3,20 @@
...
-    attr_reader :download_url
+    attr_reader :version, :intended_path, :download_url
 
     def initialize(semantic_version, intended_path, download_url = '')
       @semantic_version = semantic_version
       @intended_path = intended_path
       @download_url = download_url
-      @version = if download_url.empty?
-                   latest_xcode_revision(Xcode::Version.new(semantic_version))
+      @version = -> { determine_version }.call
```
jweyer8, line 12: Possible alternative to the lamda:

`@version = download_url.empty? ? latest_xcode_revision(Xcode::Version.new(@semantic_version)) : @samantic_version`

jweyer8 (commented):
Looks good!

jweyer8 (approved):

libraries/xcode.rb
jazaval: The lambda allows the `xcode` object to be instantiated without authenticating with Apple to see what versions are available, and allows us to first check things like `xcode.installed?` for the resource guards. The ternary would still be evaluated at initialization, which will require auth with Apple right away.

mjmerin (approved):
LGTM

Issue #244, opened 2021-12-21
[BUG] Certificate installation fails (due to wrong order of parameters?)


phirk:
## Describe the Bug 

Trying to install certificates fails:

```
certificate 'Apple Worldwide Developer Relations Certificate Authority certificate' do
  certfile "#{certificates_path}/AppleWWDRCAG3.cer"
  keychain '/Users/phirk/Library/Keychains/login.keychain-db'
  apps ['/usr/bin/security', '/usr/bin/codesign']
end
```

## Error Message

```
    Mixlib::ShellOut::ShellCommandFailed
    ------------------------------------
    execute[install-certificate] (/Users/phirk/.chef/cache/cookbooks/macos/resources/certificate.rb line 23) had an error: Mixlib::ShellOut::ShellCommandFailed: Expected process to exit with [0], but received '1'
    ---- Begin output of ["/usr/bin/security", "add-certificates", "/Users/phirk/certificates/AppleWWDRCAG3.cer", "-k", "/Users/phirk/Library/Keychains/login.keychain-db"] ----
    STDOUT:
    STDERR: security: SecCertificateAddToKeychain: A default keychain could not be found.
    security: open -k: No such file or directory
    security: SecCertificateCreateFromData: Unknown format in import.
    ---- End output of ["/usr/bin/security", "add-certificates", "/Users/phirk/certificates/AppleWWDRCAG3.cer", "-k", "/Users/phirk/Library/Keychains/login.keychain-db"] ----
    Ran ["/usr/bin/security", "add-certificates", "/Users/phirk/certificates/AppleWWDRCAG3.cer", "-k", "/Users/phirk/Library/Keychains/login.keychain-db"] returned 1
```

## Chef Exception  

`Error executing action `run` on resource 'execute[install-certificate]'`

## Observation

It seems that the order `add-certificates` *certificate* `-k` *keychain* is wrong, as it fails on the command line too:

```
sh-3.2# /usr/bin/security add-certificates /Users/phirk/certificates/AppleWWDRCAG3.cer -k /Users/phirk/Library/Keychains/login.keychain-db
security: SecCertificateAddToKeychain: A default keychain could not be found.
security: open -k: No such file or directory
security: SecCertificateCreateFromData: Unknown format in import.
```

However, the swapped order `add-certificates` `-k` *keychain* *certificate* does seem to work:

```
[...]

Issue #197, opened 2019-05-02
macos_user ignores certain sysadminctl errors, no error raised in Chef run


hawksight:
## Describe the Bug 
When you create a `macos_user` using the resource from this cookbook, the actual user creation may well fail, but Chef seems to think it has succeeded anyway because the command has run.

For example, if you have configuration to create a user, but the password does not meet the requirements, the command fails. However Chef reports back green and continues with the run.

This is problematic as it lead me to debugging the wrong step of my cookbook, the step after which depended on the success of the previous user creation.

## Error Message  

An example error message when the password does not meet requirements:

```
2019-05-02 13:40:31.419 sysadminctl[2801:17519] New account password error.  (5402)
```

## Chef Exception  

Looking in the debug logs of a chef-client run:

```
* execute[add user auser] action run[2019-05-02T13:36:07+01:00] INFO: Processing execute[add user auser] action run (/var/chef/cache/cookbooks/macos/resources/macos_user.rb line 79)

  [execute] 2019-05-02 13:36:08.057 sysadminctl[2180:14017] New account password error.  (5402)
[2019-05-02T13:36:08+01:00] INFO: execute[add user auser] ran successfully
  - execute ["/usr/sbin/sysadminctl", "-adminUser", "admin", "-adminPassword", "redacted", "-addUser", "auser", "-fullName", "Another User", "-password", "redacted", ""]
```

If you run manually the problem is probably a return code of `0` as shown below:
```
IMAC4:~ admin$ sudo /usr/sbin/sysadminctl -adminUser admin -adminPassword redacted -addUser auser -fullName "Another User" -password "redacted"
2019-05-02 13:40:31.419 sysadminctl[2801:17519] New account password error.  (5402)
IMAC4:~ admin$ echo $?
0
```

A good way to recreate is to use `green` in a password because it defies the consecutive character rule apple have. I guess this only applies when you have no set a custom password policy.

Would there be any way to add validation to the resource after an account is created?

americanhanko:
Thanks for the bug submission, @hawksight. 

Can you give us some more details about the structure of the password you're passing or the password policy you're using? Are you AD joined or do you have FileVault enabled?

hawksight:
Sure. The machine is managed initially by fleetsmith. Filevault is enabled and Firewall.

The password policy setup and enforced with Fleetsmith is as per the screenshot currently.
[...]

Issue #239, opened 2021-10-14
[BUG] defaults resource passes array values incorrectly to /usr/bin/defaults


aaronclarke:
## Describe the Bug 
defaults resource passes array values incorrectly to /usr/bin/defaults

## Error Message  
```
           ---- Begin output of /usr/bin/defaults write com.apple.dock persistent-apps -array [] ----
           STDOUT: 
           STDERR: 2021-10-14 12:55:14.087 defaults[26159:131114] Could not parse: [].  Try single-quoting it.
           ---- End output of /usr/bin/defaults write com.apple.dock persistent-apps -array [] ----
           Ran /usr/bin/defaults write com.apple.dock persistent-apps -array [] returned 1
```

As far as I can tell, the defaults resource mishandles any settings that are supposed to be arrays because it leaves the `[` and `]` characters surrounding the text representation of the value when sending it to `/usr/bin/defaults`

For example, for this defaults resource:
```ruby
defaults "com.apple.dock" do
  settings 'persistent-apps' => []
  user 'someuser'
end
```
I would expect it to end up running this command to write an empty array:
- `/usr/bin/defaults write com.apple.dock -array`

Instead it ends up running:
- `/usr/bin/defaults write com.apple.dock -array []`
- > Could not parse: [].  Try single-quoting it.

Similarly for non-empty arrays, such as:
```ruby
defaults "com.apple.dock" do
  settings 'persistent-apps' => ['dummy data']
  user 'someuser'
end
```

I would expect it to end up running:
- `/usr/bin/defaults write com.apple.dock -array 'dummy data'`

Instead it ends up running:
- `/usr/bin/defaults write com.apple.dock -array ['dummy data']`

PR #260, opened 2022-03-29
Release/5.0.0

jazaval:
### Fixed

- Extracted authentication with Apple via `xcode` resource away `xcode` object instantiation, resolving #234.
- Enabled `macos_user` resource to parse `sysadminctl` stderr, resolving #197.
- Reversed order of arguments for certificate installation, resolving #244.

### Added

- Added `apple_id` property to `xcode` resource to remove dependency on attributes or data bags for authentication.
- New certificate resource property: `kc_passwd` which allows setting of keychain password.
- New test suites and recipe change to account for `.cer` files.
- Check for certificate existence within the keychain before installing a new one to ensure idempotency.
- Made certificate password properties sensitive.
- Support for Mac Studio in `FormFactor` class. 
- Secure token support for `macos_user` resource via new properties `secure_token` and `existing_token_auth`.
- New unit and integration tests for `macos_user` resource.
- Updated our README to include Monterey support. 
- Updated certificate resource documentation.
- Added the documentation directory to chefignore as we don't need to upload all our docs to the Chef Infra Server. 

### Changed

- Deprecated `plist` resource in favor of the `plist` resource included with Chef Client >=16.
- Unified `macos_user` test suites.
- Updated `macos_user` resource to use not utilize default attributes for authorization.
- Updated all deprecated tool names, shell outputs, and URLs in `TESTING.md`

### Removed

- Removed the ability to authenticate with Apple for `xcode` downloads via node attributes or data bags.
- Removed dependency on using the `default['macos']['admin_password']` attribute for setting the keychain password when using the certificate resource.
- Removed last default cookbook attributes:
  - `node['macos']['admin_user']`
  - `node['macos']['admin_password']`
  - `node['macos']['apple_id']`


rodreag (commented):
The list of desktop Macs should also include 'Mac' which is what sysctl hw.model returns for Mac Studios:
Mac13,1

README.md
```diff
@@ -1,49 +1,34 @@
...
-Each of these attributes defaults to vagrant since our resources are developed
-with the Vagrant paradigm. In other words, the user and password declared here
-should be an admin user with passwordless super-user rights.
+- macOS 10.14 Mojave
+- macOS 10.15 Catalina
+- macOS 11 Big Sur
+- macos 12 Monterey
 
 ## Resources
+**The plist resource is deprecated and will be removed in the next major release of macos-cookbook. Please use the plist resource included with Chef Client >=16.**
```
aaronclarke, line 24: Do we know if the Chef Client version of the plist resource makes use of `defaults` in order to make sure the defaults cache is updated correctly?
I couldn't find any mention of `cfprefsd` or the `defaults` utility on Chef's page:
https://docs.chef.io/resources/plist/

aaronclarke, line 24: The macos-cookbook's implementation of `plist` supports the ability to specify arrays, but the Chef Client implementation does not:
> value
> Ruby Type: true, false, String, Integer, Float, Hash

https://docs.chef.io/resources/plist/

jazaval: It does not, but neither does the macos `plist`. There are no plans to deprecate the `defaults` resource or any of the libraries it depends on.

jazaval: That's odd - it should since it was ported directly from this cookbook. It seems like functionality is there https://github.com/chef/chef/blob/a7b347ceb49f66ee5797ae578fbbcc4e45983aee/lib/chef/resource/plist.rb#L164, but is not listed in the value types.

jazaval: I'm not sure the macos implementation explicitly supports arrays either: https://github.com/microsoft/macos-cookbook/blob/103fab4446c74dde8cf8412e936e1584f59d09f5/resources/plist.rb#L7

rodreag (approved):
Looks good

README.md
aaronclarke: Your mention in the other comment thread about `defaults` resource made me realize that I may be getting the two resources confused. Certainly the issue I had previously filed about support for arrays was for the `defaults` resource, so that's probably it.
#239

libraries/developer_account.rb
```diff
@@ -5,32 +5,35 @@ module MacOS
...
       end
     end
 
-    def authenticate_with_apple(credentials)
-      shell_out!(XCVersion.update, env: credentials)
-    end
-
-    def find_apple_id(data_bag_retrieval, node_credential_attributes)
-      if node_credential_attributes
+    def validate_apple_id(apple_id_hash)
```
americanhanko, line 19: Instead of the duplication here, maybe try to catch a `KeyError` using `Hash#fetch`:

```ruby
if apple_id_hash.values.any?
    {
      username: apple_id_hash[:username],
      password: apple_id_hash[:password],
     }
else
    begin
        {
          username: ENV.fetch('XCODE_INSTALL_USER'),
          password: ENV.fetch('XCODE_INSTALL_PASSWORD'),
        }
    rescue KeyError
        raise('Developer credentials not supplied, and a URL was not provided for Xcode!')
    end
end
```

Also, not sure why the error message says a URL was not provided for Xcode? This looks like it's only related to user credentials.

libraries/system.rb
```diff
@@ -9,12 +9,12 @@ def initialize(hardware)
 
       def desktop?
         return false if @machine_model.nil?
-        @machine_model.match? Regexp.union %w(Macmini MacPro iMac)
+        @machine_model.match? Regexp.union ['Macmini', 'MacPro', 'iMac', /Mac\d/]
```
americanhanko, line 12: I feel like there's a better way to determine whether or not a system is laptop or a desktop. Maybe we could check for a built-in display or a battery or both?

libraries/xcode.rb
```diff
@@ -3,18 +3,20 @@
...
     def initialize(semantic_version, intended_path, download_url = '')
       @semantic_version = semantic_version
       @intended_path = intended_path
       @download_url = download_url
-      @version = if download_url.empty?
-                   latest_xcode_revision(Xcode::Version.new(semantic_version))
+      @version = -> { determine_version }
+    end
+
+    def determine_version
```
americanhanko, line 15: Is there a test for this newly added logic? It might be less obvious since the method doesn't have any parameters.

resources/certificate.rb
```diff
@@ -1,26 +1,31 @@
...
-property :keychain, String
-property :apps, Array
-
-action_class do
-  def keychain
-    new_resource.property_is_set?(:keychain) ? new_resource.keychain : ''
-  end
-end
+property :cert_password, String, sensitive: true
+property :keychain, String, required: true
```
americanhanko, line 8: We changed the property in the `keychain` resource to `kc_file`, but we're keeping leaving it as `keychain` for the certificate resource? Personally, I think it would be better to have the properties align if they're referring to the same kind of file. Particularly important since it accepts `kc_passwd`.

And while we're on the topic, is there a reason we truncated these properties? `keychain_file` and `keychain_password` would be more preferable all across the board IMHO, for readability and clarity.

test/cookbooks/macos_test/recipes/certificate.rb
```diff
@@ -1,12 +1,42 @@
...
+end
+
+openssl_x509_certificate foobar_pem_path do
+  common_name 'www.f00bar.com'
+  org 'Foo Bar'
+  org_unit 'Lab'
+  country 'US'
+end
+
+execute 'convert .pem certificate to .cer certificate' do
```
americanhanko, line 22:
```suggestion
execute 'convert certificate to binary format' do
```

test/cookbooks/macos_test/recipes/certificate.rb
```diff
@@ -1,12 +1,42 @@
...
+  command ['/usr/bin/openssl', 'x509', '-inform', 'PEM', '-in', foobar_pem_path, '-outform', 'DER', '-out', foobar_cer_path]
+  only_if { ::File.exist? foobar_pem_path }
+end
+
+certificate 'install a .cer format certificate file' do
```
americanhanko, line 27:
```suggestion
certificate 'install a binary-formatted certificate' do
```

test/cookbooks/macos_test/recipes/xcode_from_url.rb
```diff
@@ -1,11 +1,11 @@
-if node['platform_version'] >= '10.14.4'
-  xcode 'installs 11.0' do
-    download_url node['xcode']['download_url']
-    version '11.0'
+if node['platform_version'] >= '10.15.6'
+  xcode 'installs 12.0' do
+    download_url ::File.read('/Users/vagrant/xcode')
```
americanhanko, line 3: Why do we need to get the URL from a file?

test/integration/default/controls/security_test.rb
```diff
@@ -11,9 +11,17 @@
...
-  describe command('/usr/bin/security find-certificate /Users/vagrant/Library/Keychains/login.keychain') do
+  describe file('/Users/vagrant/foobar.cer') do
+    it { should exist }
+  end
+
+  describe command('/usr/bin/security find-certificate /Users/vagrant/Library/Keychains/test.keychain') do
     its('stdout') { should include 'Test' }
   end
+
+  describe command('/usr/bin/security find-certificate /Users/vagrant/Library/Keychains/login.keychain') do
```
americanhanko, line 22: I think the output of this command only lists the first certificate found. Maybe it would be better to use `-n` (query by name) or `-e` (query by email) to be more specific. Even better would be to use `-p` to print the contents in PEM format and verify the contents exactly match those of the original PEM file that we installed before.

americanhanko (approved):
Approved with a handful of suggestions, with the most important being the discrepancy between the `keychain` and `kc_file` properties for the `certificate` and `keychain` resources, respectively.

libraries/xcode.rb
jazaval: `determine_version` is just an extraction of the assignment that's from `initialize`. We're covered here with the tests that instantiate Xcode with the `download_url` parameter.

TESTING.md
```diff
@@ -34,8 +33,8 @@ business logic for each of the custom resources and are used as either mixins
 for the custom resources or contain classes that act as helpers, except with
 explicit namespacing. Some libary files are heavily unit tested, others are definitely
 missing much-needed unit test coverage. The following command syntax assumes
-you've made the ChefDK `rspec` your default by running `chef shell-init bash`
-(https://docs.chef.io/ctl_chef.html#chef-shell-init)
+you've made the Chef Workstation `rspec` your default by running `chef shell-init bash`
```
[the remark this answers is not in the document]
v-anshie: When are we actually making the switch over to `zsh` since Apple pushes it as 100% default now? I know we had a long conversation years ago about this but can't remember if we planned to make a change in the future for it or not.

test/cookbooks/macos_test/recipes/certificate.rb
jazaval: Added to `feature/keychain`.

resources/certificate.rb
jazaval: Updating in `feature/keychain`.

test/cookbooks/macos_test/recipes/certificate.rb
jazaval: Added to `feature/keychain`.

TESTING.md
```diff
@@ -85,29 +84,49 @@ a few issues that need to be addressed before doing so.
...
+microsoft/macos-high-sierra (parallels, 10.13.6)
+microsoft/macos-mojave      (parallels, 10.14.6)
+microsoft/macos-monterey    (parallels, 12.2)
 ```
 
 Next, make sure you're in the macOS cookbook root and run `kitchen list` to view
 the available instances. It should look something like this:
 
 ```shell
 $ kitchen list
```
v-anshie, line 103: This isn't a huge deal, but if we already pin the version in `kitchen.yml` would we need to put the chef version in the `platform` names? I assume its just a visibility thing but it looks like it clutters IMHO.

test/cookbooks/macos_test/recipes/xcode_from_url.rb
jazaval: Kitchen lifecycle doesn't seem to have support for porting environment variables, and I didn't want to hardcode this. Previously, we set this via a node attribute in kitchen (`node['xcode']['download_url']`), and I wanted to be able to entirely delete the default attributes file.

v-anshie (approved):

libraries/developer_account.rb
jazaval: Not sure which duplication this resolves, you have two sets of potential assignment to the `username` and `password` keys. I actually think the `if`/`elsif`/`else` is more straightforward than nesting `begin`/`rescue` under `else`.

You don't have to supply developer credentials if there is a URL for an Xcode xip provided. This raises if neither was provided.

TESTING.md
jazaval: This is leftover from when we tested two different versions of Chef - I'll leave it since we'll probably test 18 and 17 at the same time for a bit.

libraries/system.rb
jazaval: This does feel a little less elegant now that the regex isn't as clean, but I think the frequency at which new machine models are released is rare enough that this is okay.

test/integration/default/controls/security_test.rb
jazaval: The PEM is generated during the test recipe execution, so we can't hardcode it into the integration test. I've added the arguments to query all certificates in the keychain.

PR #262, opened 2022-04-14
align certificate and keychain property names, refactor SecurityCommand module

jazaval:
- Changed `certificate` property names to be more clear within resource scope and consistent with `keychain` resource:
  - `certfile` is now `path`
  - `cert_passwd` is now `password`
  - `keychain` is now `keychain_path`
- Changed `keychain` property names to be more clear within resource scope and consistent with `certificate` resource:
  - `kc_file` is now `path`
  - `kc_passwd` is now `password`
 - Refactored SecurityCommand module to use named parameters instead of empty strings for optional instantiation arguments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,19 +11,26 @@
 ### Added
 
 - Added `apple_id` property to `xcode` resource to remove dependency on attributes or data bags for authentication.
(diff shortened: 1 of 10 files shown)


rodreag (approved):
Looks good

jweyer8 (approved):

Hypervividdream (approved):

PR #264, opened 2022-05-23
Fix autologin for 11.6.6

sjkunkler:
Add setup assistant keys for 11.6.6 OS update

Fixed `macos_user` resource autologin functionality after 11.6.6 update.


sjkunkler:
I'm assuming you'll want to update the version and add to the change log when you create the release branch, but if not let me know!

jazaval (approved):

PR #270, opened 2022-11-16
Better check for if Xcode is properly installed

GuMiner:
## Problem
If Xcode install fails after the download but before Xcode is completely installed, subsequent Chef runs won't properly reinstall Xcode.

## Change  
This change adds an additional check to the `installed?` definition, ensuring that `/usr/bin/xcodebuild -showsdks` returns an exit code of 0 to consider Xcode properly installed.

### DIFF
--- a/libraries/xcode.rb
+++ b/libraries/xcode.rb
@@ -50,7 +50,7 @@ module MacOS
     def installed?
       return false if installed_path.nil?
 
-      installed_path.any? && shell_out("DEVELOPER_DIR='#{installed_path}' /usr/bin/xcodebuild -showsdks").exitstatus == 0
+      installed_path.any? && shell_out("DEVELOPER_DIR='#{current_path}' /usr/bin/xcodebuild -showsdks").exitstatus == 0
     end
 
     def compatible_with_platform?(macos_version)
--- a/libraries/xcode.rb
+++ b/libraries/xcode.rb
@@ -50,7 +50,7 @@ module MacOS
     def installed?
       return false if installed_path.nil?
 
-      installed_path.any? && shell_out("DEVELOPER_DIR=#{installed_path} /usr/bin/xcodebuild -showsdks").exitstatus == 0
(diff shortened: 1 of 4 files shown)


GuMiner:
FYI this'll fail the `chef exec rspec` check until #269 is merged to fix that check.

jazaval (approved):

libraries/xcode.rb
```diff
@@ -50,7 +50,7 @@ def current_path
     def installed?
       return false if installed_path.nil?
 
-      installed_path.any?
+      installed_path.any? && shell_out('/usr/bin/xcodebuild -showsdks').exitstatus == 0
```
aaronclarke, line 53: /usr/bin/xcodebuild is not a perfect standin for the health of Xcode.
1) xcodebuild will run against whatever the developer dir is / whatever `xcode-select --print-path` is pointing at. If that isn't pointing at your new Xcode, you'll get the wrong result.
2) This doesn't handle the situation that someone is installing multiple builds of Xcode because you can have more than one Xcode but you'll only have one `xcodebuild`. Not sure if that's a common case

aaronclarke:
```suggestion
      installed_path.any? && shell_out("DEVELOPER_DIR=#{installed_path} /usr/bin/xcodebuild -showsdks").exitstatus == 0
```

libraries/xcode.rb
```diff
@@ -50,7 +50,7 @@ def current_path
     def installed?
       return false if installed_path.nil?
 
-      installed_path.any?
+      installed_path.any? && shell_out("DEVELOPER_DIR=#{installed_path} /usr/bin/xcodebuild -showsdks").exitstatus == 0
```
aaronclarke, line 53:
```suggestion
      installed_path.any? && shell_out("DEVELOPER_DIR='#{installed_path}' /usr/bin/xcodebuild -showsdks").exitstatus == 0
```

aaronclarke (approved):

jazaval (approved):

rodreag (approved):
Looks good

jweyer8 (approved):
👍

ehanlon (approved):

PR #274, opened 2023-06-20
Be able to install command line tools if machine previously installed them but were wiped from a major OS upgrade

hvongsachang:
After a major macOS upgrade, command line tools are usually removed. However, the /Library/Receipts/InstallHistory.plist file still keeps history of the command line tools that were installed, so the machine thinks that they are installed.

This causes the machine to go through these steps:
1. '/Library/Developer/CommandLineTools/usr/lib/libxcrun.dylib' does not exist (because of the os upgrade), so it executes the install block
2. The version it tries to install is the latest installed command line tools according to InstallHistory.plist.
3. We don't try to scan the softwareupdate catalog because the InstallHistory.plist has previously recorded CLT installs.
4. Since we don't scan the catalog, we never create the sentinel file.
5. Because there's no sentinel file, we can't install the CLT that was on the machine prior to macOS upgrade.

This problem is easily fixed if we touch the sentinel file prior to install. This change slightly refactors where we touch the sentinel file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## [5.0.5] - 2023-06-20
+
+### Fixed
+- Updated `command_line_tools` so that machines are able to install previously installed command line tools if they were wiped from a major macOS upgrade.
+
 ## [5.0.4] - 2023-01-31
 
 ### Added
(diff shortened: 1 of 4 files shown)


GuMiner:
Also update the CHANGELOG too for the new patch version: 
https://github.com/microsoft/macos-cookbook/blob/master/CHANGELOG.md

GuMiner (approved):

ehanlon (approved):

PR #278, opened 2023-06-27
Release/5.1.0

jazaval:
# Changelog

### Fixed

- Fixed issues with ARD not working on macOS Monterey.
- Ensured that CLT demand file always exists before we query the SWU catalog.

### Added

- Added additional functionality to the [remote_management](resources/remote_management.rb) resource.
  - You can now specify the users to whose privileges will be configured.
  - You can now specify the privileges to bestow upon the given users.
  - You can now set the computer info fields; this is helpful for stratifying computers within ARD.


jazaval:
@jweyer8 can you take a look at the *.bak file from the `remote_management_updates` branch?

kitchen.yml
```diff
@@ -19,33 +20,20 @@ verifier:
...
-- name: monterey-chef17
+- name: monterey-x86
   driver:
     box: microsoft/macos-monterey
-    box_version: 12.6.3
-  provisioner:
-    product_version: 17
+    box_version: 12.6.7
+
+- name: ventura-arm
```
jweyer8, line 33: Do we want to test ventura-x86 as well?

jazaval: Still working on getting those boxes working on our Ventura hosts before we open those floodgates, but yea we can add that suite and leave it off of the pipelines.yml

resources/command_line_tools.rb
```diff
@@ -2,45 +2,48 @@
...
 
 action :install do
-  command_line_tools = CommandLineTools.new
-
-  file 'create sentinel file' do
+  file 'create demand file' do
     path '/tmp/.com.apple.dt.CommandLineTools.installondemand.in-progress'
     group 'wheel'
+    not_if { ::File.exist?('/Library/Developer/CommandLineTools/usr/lib/libxcrun.dylib') }
+    notifies :run, 'execute[install command line tools]', :immediately
```
hvongsachang, line 15: So if we tell the `execute[install command line tools]` block to run immediately, would that skip over line 18 where we create a new `CommandLineTools` object?

jazaval: it won't ship - the execution path when both "Chef DSL" and Ruby co-exist is confusing. TL;DR is that `immediately` means not `delayed` until the end of the Chef run: https://docs.chef.io/resource_common/#notifications

hvongsachang (approved):
Approving for the changes made related to command line tools!

resources/remote_management.rb
```diff
@@ -3,24 +3,85 @@
...
+         coerce: ->(p) { p.is_a?(Array) ? p.compact.map(&:to_s) : [p] },
+         callbacks: { 'has too many elements; computer info excepts up to four info fields' => ->(p) { (p.is_a?(Array) && p.size < 4) } }
+
+load_current_value do |desired|
+  current_value_does_not_exist! unless RemoteManagement.activated? && RemoteManagement.current_users_configured?(desired.users) && RemoteManagement.current_users_have_identical_masks?(desired.users)
+  privileges RemoteManagement.current_mask(desired.users)
+  computer_info RemoteManagement.current_computer_info
+end
+
+# TODO; the enable action should be decoupled from configuration; configure action should be added; default action should be [:configure, :enable]
```
jweyer8, line 31: @jazaval what do you think about splitting the `:enable` action into two actions `[:configure, :enable]`. `:enable` is a bit misleading as a lot of configuration is done.

jweyer8 (approved):

PR #282, opened 2023-10-17
User/mswartz/add latest os

Matthew-Swartz:
Adding Ventura and Sonoma platforms ~(linked_clone to be removed later when Jacob finishes his work)~ fixed


jazaval (approved):
LGTM

PR #285, opened 2024-04-04
Release/6.0.0

jazaval:
### Removed

- Removed the `plist` resource

### Fixed

- Fixed issues with ARD not working on macOS Monterey.
- Fixed an xcode-install gem issue.

### Added

- Added additional functionality to the [remote_management](resources/remote_management.rb) resource.
  - You can now specify the users to whose privileges will be configured.
  - You can now specify the privileges to bestow upon the given users.
  - You can now set the computer info fields; this is helpful for stratifying computers within ARD.

- Added support for `macos_user` to create an account with no password.
- Added support for `command_line_tools` to install beta versions.


azure-pipelines.yml
```diff
@@ -62,18 +61,18 @@ jobs:
...
   parameters:
-    jobName: "vintage"
-    platforms:
-    - big-sur-x86
-    - monterey-x86
-    suites: ${{ parameters.kitchenSuites }}
+    jobName: "vintage_agent"
     kitchenFile: kitchen.yml
+    platforms:
+      - big-sur-x86
```
GuMiner, line 76: BigSur is no longer officially supported -- does testing still need to be done on that platform?

kitchen.yml
```diff
@@ -23,36 +23,28 @@ platforms:
 - name: big-sur-x86
```
GuMiner, line 23: Here's another BigSur instance that potentially could also go away.

libraries/macos_user.rb
```diff
@@ -1,9 +1,13 @@
...
     def kcpassword_hash(password)
-      bits = magic_bits
+      bits = if password.nil?
+               nil_magic_bits
+             else
+               magic_bits
+             end
       obfuscated = []
-      padded(password).each do |char|
+      padded(password.to_s).each do |char|
```
GuMiner, line 10: As far as I know `.to_s` on `nil` in Ruby returns `""`. This effectively does nothing if `password.nil?` is true.

resources/macos_user.rb
```diff
@@ -73,13 +73,26 @@ def user_fullname
...
+      ''
+    end
+  end
+
   def exec_sysadminctl(args)
     shell_out!(sysadminctl, args).stderr
   end
 
+  def logged_in?(user)
+    logged_in_user = shell_out('stat', '-f', '%Su', '/dev/console').stdout.chomp
```
GuMiner, line 89: This appears to work, although I'm wondering why `stat` was used instead of `whoami`?

GuMiner (approved):
Left a few comments, but nothing that seems overly problematic. We might need to make changes on our end to accommodate the `plist` deprecation, but if necessary we should have done that a while ago.

libraries/macos_user.rb
jazaval: Ruby raises an exception that there is no *implicit* conversion of nil to String without this.

resources/macos_user.rb
jazaval: `whoami` returns the user running in the tty, vs `stat /dev/console` returns the user with an active macOS UI session

sjkunkler (approved):

PR #290, opened 2025-03-26
Allow ability to disable security response updates in macOS 15 (Sequoia)

iamjessee:
---
name: Pull Request
about: File a pull request to help us improve
title: "[PR]"
labels: ''
assignees: ''

---

## Describe the Pull Request  
These changes are being added to allow users of this cookbook to also disable security responses on macOS 15 (Sequoia) machines.

## Justification  
This change is being made due to when using this cookbook on machines running macOS 15 (Sequoia) or later I noticed that Security Response is still enabled on machines even when all current values are set to false. I found that on older version of macOS the current implementation works fine, however on macOS 15 (Sequoia) I had to add manually disable this feature in the GUI or add my own method to the recipe using this cookbook to disable this myself. It looks like in older version of macOS the option 'ConfigDataInstall' was tied to 'AutomaticDownload', so when that value was set to FALSE it would follow suite. Now it appears this has been changed to be independent and must be editing on its own to disable Security Response updates.


iamjessee:
@microsoft-github-policy-service agree

jazaval (approved):

PR #308, opened 2026-06-04
Install specific version of signet to mitigate Ruby 3.2+ requirement

dsiher:
---
name: Install specific version of signet to mitigate Ruby 3.2+ requirement
---

The latest release of signet requires Ruby 3.2+ which both Chef and CINC 18.x do not support. We can undo this when we eventually update to 19.x in the future.


jazaval (approved):
LGTM

